PR #1, opened 2015-12-29
add objective, Adrian, and alphabetize

bensigelman:


bensigelman:
@yurishkuro I think (?) that if I merge this the autogen web stuff will Just Work. Let me know if you have any proposed changes to the copy. (I rolled back my direct html edit)

yurishkuro:
I think if you edit json directly, github doesn't regenerate HTML. So I went to page builder again and re-saved, which made the changes visible.

Issue #8, opened 2016-01-07
document semantic API 1.0


bensigelman:
Add a concise description of the semantic requirements for per-platform opentracing APIs to the static site.

bensigelman:
sub-bullet: standardize what's already there (`Reporter` vs `Recorder`, e.g.)

bensigelman:
I merged this so will close the issue, though we're not "done" done since I think there will be a few actual semantic changes to the APIs.

PR #13, opened 2016-01-10
Clarifies standard api types for span

adriancole:
Prior to this change, there was a mutable accumulator called `Span` (similar to HTrace's one), and also a concept of a immutable finished span called `RawSpan`. Since `RawSpan` was the input of `RecordSpan`, and "raw" distracts at least me, this changes terminology to say `SpanRecord`. I believe `SpanRecord` more closely relates both the the apis that use it, and also it being the canonical immutable struct.

### DIFF
--- a/apis.md
+++ b/apis.md
@@ -9,3 +9,10 @@ The following OpenTracing APIs are currently available in the RFC stage:
 
 Please refer to the README files in the respective projects for examples of usage.
 
+### Implementation Notes
+
+OpenTracing APIs intend to express a similar experience to the user, while still being idiomatic to the language or framework. The following concepts apply to all variants.
+
+A **Span** is shared mutable state. It accumulates tags and logs, and can also create children. Finishing a span means capturing its state as a SpanRecord.
+
+A **SpanRecord** is an immutable representation of a completed span. It can be reported directly to an OpenTracing system, or encoded into a representation of an alternate system, such as HTrace or Zipkin.


apis.md
```diff
@@ -9,3 +9,10 @@ The following OpenTracing APIs are currently available in the RFC stage:
 
 Please refer to the README files in the respective projects for examples of usage.
 
+### Implementation Notes
+
+OpenTracing APIs intend to express a similar experience to the user, while still being idiomatic to the language or framework. The following concepts apply to all variants.
+
+A **Span** is shared mutable state. It accumulates tags and logs, and can also create children. Finishing a span means capturing its state as a SpanRecord.
+
+A **SpanRecord** is an immutable representation of a completed span. It can be reported directly to an OpenTracing system, or encoded into a representation of an alternate system, such as HTrace or Zipkin.
```
yurishkuro: nit: these details might be better placed in the terminology page, or perhaps in a future Architecture page where we can describe the relationships between Construction & Recording layers. Note that SpanRecord is only meaningful for tracers that extend standard tracer, and uses of OpenTracing API should never need to know about SpanRecord.

adriancole: >
> nit: these details might be better placed in the terminology page, or
> perhaps in a future Architecture page where we can describe the
> relationships between Construction & Recording layers.
>
good point.. I started this in the terminology page, but didn't want to
distract users (or scare them away!) with this sort of nuance.

> Note that SpanRecord is only meaningful for tracers that extend standard
> tracer, and uses of OpenTracing API should never need to know about
> SpanRecord.
>
Good point.. I missed that. Maybe this should be a doc about "standard
tracer"

adriancole:
this is more about #8

apis.md
bensigelman: For the v1.0 API I'm less and less confident that we should really talk about the "standard tracer" at all here... it's only interesting to the tracing system implementors, and in terms of cardinality there are hopefully going to be far more "instrumentors" than implementors. Perhaps there should be a separate area of the docs for tracing system implementors, and that's where things like this could be discussed. Thoughts?

yurishkuro: I agree, implementations of the API should be discussed in a separate area.

> For the v1.0 API I'm less and less confident that we should really talk about the "standard tracer"

Do you still think we should provide one in the API projects?  I got distracted today by the API harness (which already paid off once I ran our real impl against it), so haven't started on the standard tracer for Python.  Perhaps instead of doing that I could fix the `zipkin-like` to be testable. We could also move `zipkin-like` out into its own repo, and actually call it `python-zipkin`, as in the "reference implementation". Problem is it doesn't have a real reporter to submit to Zipkin.

adriancole: >
> I agree, implementations of the API should be discussed in a separate area.
>
> I third the notion!

> For the v1.0 API I'm less and less confident that we should really talk
> about the "standard tracer"
>
> Do you still think we should provide one in the API projects? I got
> distracted today by the API harness (which already paid off once I ran our
> real impl against it), so haven't started on the standard tracer for
> Python.
>
I'm not sure, but seems if it were called Mock or FakeTracer, it would
still allow folks to playground, and write test harnesses, but without the
warranties.

> Perhaps instead of doing that I could fix the zipkin-like to be testable.
> We could also move zipkin-like out into its own repo, and actually call
> it python-zipkin, as in the "reference implementation". Problem is it
> doesn't have a real reporter to submit to Zipkin.
>
Other tracers for zipkin exist in the OpenZipkin org, and have coupling
there. I think it would make more sense for a new python zipkin tracer to
exist as "zipkin-python" or similar there.. @abesto thoughts?

abesto: I'm slightly out of the loop on this, trying to piece together things, here's one setup that would make sense to me. Feel free to point out if I'm missing the real conversation.

`opentracing/api-python` implements the standard API in a way that makes it relatively easy to plug in "real" implementations, and provides a no-op implementation (by default I'd call it `NoopTracer`, but consistency with the other API implementations is more important here).

`openzipkin/zipkin-python` (or `openzipkin/zipkin-traces-python`, because we may want to build an query API client too and other kinds of python thingies) implements the same API, preferably by plugging into `opentracing/api-python`, making it possible for users to switch the Zipkin reporter to whatever else while keeping their application code unchanged.

For this to be possible the import path needs to not change when swapping out the backend. Two ways to do that would be stuff like `opentracing/api-django` which uses the facilities of the Django framework to do exactly this (replicate for all "supported" frameworks); or setting up light-weight classloader-like dynamic import functionality in `opentracing-api-python`. The former is more specific, while the latter is more surprising.

There's a Zipkin python reporter that should be easy to generalize (see https://github.com/prezi/django-zipkin/issues/8), with the caveat that it doesn't adhere to the standard API. This can be a good thing in that it provides an opportunity to test the plugin-like architecture, if we go that way.

Whatever approach we choose, we'll probably want to provide integration with common frameworks, especially when it comes to configuring endpoints for various reporters.

abesto: Oh, one more thing: deeper framework integration also makes sense; for example, entering a singleton tracing context in the pre-request middleware and providing easy access to it is something I'd be looking for.

yurishkuro: @abesto I have a repo that integrates OpenTracing API with a number of other frameworks in Python, I am just waiting on the legal to make the repo on github private, eta this week. By using that instrumentation package and our internal  plug-in implementation of the tracer, enabling tracing in Python apps is reduced to a configuration change.

yurishkuro: > I'm not sure, but seems if it were called Mock or FakeTracer, it would still allow folks to playground, and write test harnesses, but without the warranties.

@adriancole  Can you elaborate which features you would want to see in such Mock tracer?  And if we have example/zipkin-like converted to a real reference implementation, wouldn't it be sufficient as a playground? I am a bit uneasy about maintaining a full-featured "fake" implementation.

> Other tracers for zipkin exist in the OpenZipkin org, and have coupling there. I think it would make more sense for a new python zipkin tracer to exist as "zipkin-python" or similar there.. @abesto thoughts?

That's fine with me: `openzipkin/opentracing-zipkin-python`

bensigelman: `openzipkin/opentracing-zipkin-python`: :+1: 

That's the "cleanest" option in my mind.

adriancole: with a notable goof (docker-zipkin), the convention has been to prefix
`zipkin` on repo names. ex. `zipkin-dependencies-spark` or
`zipkin-mongodb`. nit, but conventionally, it seems better as
`zipkin-python-opentracing`, if the intent were to have a separate repo for
the tracing component vs other python components, such as a backend.

adriancole: >
> @adriancole <https://github.com/adriancole> Can you elaborate which
> features you would want to see in such Mock tracer? And if we have
> example/zipkin-like converted to a real reference implementation, wouldn't
> it be sufficient as a playground? I am a bit uneasy about maintaining a
> full-featured "fake" implementation.
>
Fake/InMemory/Buffering tracer would have no dependencies. Ex many tracing
systems require that you start processes etc. Fake could keep state
in-memory which would allow us to define tests. This approach is quite
common. In fact, zipkin itself, both java and scala, have a in-memory span
store for the purpose of defining tests. Finagle has a BufferingTracer,
etc. This is also quite common outside tracing. For example, most large
projects I've worked on, jclouds, okhttp, etc. have a fake. Also, internal
work at square, we used fakes as ways of dealing with microservice
architecture testing when a boundary isn't in your teams control. Long
story short, is that we can make this sort of rigor optional, but I'd
certainly not write a system without a fake in 2016.

For users, I suspect a LoggingTracer would be the most useful way to
step-into the project. This again has no dependencies, so they don't need
to run docker or any other external process. Again, this is extremely
normal behavior, and for example, finagle has a console tracer.

yurishkuro: that's what I've implemented in api-go as well, InMemoryRecorder - https://github.com/opentracing/api-go/blob/master/testutils/recorder.go#L13

api-python/examples/zipkin-like has a LoggingReporter, which could use some improvement on the output format as it just calls a toString on the span. An InMemoryReporter can be added easily.

So I think zipkin-like package satisfies your reqs, but it's a real implementation, not a fake, just depends on which reporter you use it with. I didn't give it a real reporter because that depends on the collector implementation - could be http post to query svc, or save to kafka. One we move it to openzipkin repo, we'd want to add some real reporter, or two.

adriancole: and spiked!

https://github.com/openzipkin/zipkin-python-opentracing

Added Ben, Yuri and Zoltan as contributors

adriancole: >
> that's what I've implemented in api-go as well, InMemoryRecorder - https://github.com/opentracing/api-go/blob/master/testutils/recorder.go#L13
>
> yeah that looks like same concept at first glance.

> api-python/examples/zipkin-like has a LoggingReporter, which could use
> some improvement on the output format as it just calls a toString on the
> span. An InMemoryReporter can be added easily.
>
ack

> So I think zipkin-like package satisfies your reqs, but it's a real
> implementation, not a fake, just depends on which reporter you use it with.
> I didn't give it a real reporter because that depends on the collector
> implementation - could be http post to query svc, or save to kafka. One we
> move it to openzipkin repo, we'd want to add some real reporter, or two.
>
I think mostly same page. I think the zipkin-like becomes zipkin real in
the repo as you mentioned, but I do think something that defines tests
shouldn't live downstream.

IOTW, at least in api-java I plan to use the same route as we do in zipkin
for span stores. Which is to define tests and a fake that live in the
authoritative repo. Downstream implementations must pass or explicitly skip
these tests. This simplifies answer to the question of what behavior should
be vs what one or another downstream do.

bensigelman: thanks for getting https://github.com/openzipkin/zipkin-python-opentracing going, @adriancole!

yurishkuro: > IOTW, at least in api-java I plan to use the same route as we do in zipkin for span stores. Which is to define tests and a fake that live in the authoritative repo. Downstream implementations must pass or explicitly skip these tests. This simplifies answer to the question of what behavior should be vs what one or another downstream do.

@adriancole yes for langs like Java/Go/etc it makes more sense to define a fake. Not sure if you saw the test harness I added in Python, it basically exercises all API methods against a given tracer, and in the main api-python repo it just runs against the no-op tracer which itself serves as the API definition.

Unfortunately, given that the API is mostly "write", there are limited things we can do in the tests to actually check correctness of the implementation. One possibility is to define some "read" API that must be implemented by the Reporter or some anonymous component of the tracer based on the data it captured, purely for compliance testing purposes.

adriancole: >
> Unfortunately, given that the API is mostly "write", there are limited
> things we can do in the tests to actually check correctness of the
> implementation. One possibility is to define some "read" API that must be
> implemented by the Reporter or some anonymous component of the tracer based
> on the data it captured, purely for compliance testing purposes.
>
ack that a read or assertion api is needed or else the tests can't really
do any correctness checking. I think a compatibility test which doesn't
check correctness isn't complete. So, basically I'd at least put a TODO on
it, if no solution is in mind, yet.

adriancole:
I don't think there's a single path of action this issue can take. closing it in favor of other work.

Issue #9, opened 2016-01-08
finish main opentracing.io static site


bensigelman:
- [x] briefly explain the raison d'etre for tracing systems (visibility, etc)
- [x] briefly explain the raison d'etre for opentracing (flexibility, decoupling, etc)
- [x] point to language-independent "semantic spec" (#8)
- [x] point to each language-specific repo
- [x] brief notes for tracing system implementors
- [x] separate contributors page
- [x] host on opentracing.io without the redirect

yurishkuro:
> point to each specific language-specific repo

this is done under APIs page. Do you have something else in mind?

bensigelman:
yeah, fair point -- I was going to put it on the main page, too, but that's probably just going to fall out of date.

yurishkuro:
For contributors page in the future we could use something like http://blossom.github.io/contributors/

PR #15, opened 2016-01-10
give more context about the project

bensigelman:


bensigelman:
(per #9)

btw, I briefly looked into the blossom contributors thing... I ran into some headaches with custom js in jekyll and gave up for now.

yurishkuro:
lgtm

adriancole:
:shipit:

_config.yml
```diff
@@ -26,8 +28,8 @@ permalink:        pretty
 
 # Setup
 title:            The OpenTracing Project
-tagline:          'Common APIs for distributed tracing and context propagation'
-description:      'Common APIs for distributed tracing and context propagation'
+tagline:          'Safe, general APIs for distributed tracing and context propagation'
+description:      'Safe, general APIs for distributed tracing and context propagation'
```
yurishkuro: "safe" in what sense?

Would be nice the first word to emphasize the interop property of the API.

bensigelman: yeah, "safe" is the wrong word... I changed this and also added a little more detail to the "APIs" page.

bensigelman: (btw, I meant "safe" in that they wouldn't get locked in)

_config.yml
```diff
@@ -26,8 +28,8 @@ permalink:        pretty
 
 # Setup
 title:            The OpenTracing Project
-tagline:          'Common APIs for distributed tracing and context propagation'
-description:      'Common APIs for distributed tracing and context propagation'
+tagline:          'Consistent and expressive APIs for distributed tracing and context propagation'
+description:      'Consistent and expressive APIs for distributed tracing and context propagation'
```
yurishkuro: > (btw, I meant "safe" in that they wouldn't get locked in)

Vendor-neutral might be a good descriptor.

bensigelman: Sure, I'll add that to my next PR on the docs stuff.

adriancole: +1 vendor neutral

bensigelman: yeah, that fix is pending in my next round of updates... will send that out later today for sure. (yesterday was lost to non-laptop work)

Issue #10, opened 2016-01-08
refer to main site for non-platform-specific content in language READMEs


bensigelman:
(rather than let them all go stale)

slimsag:
+1 highly agree

bensigelman:
(with https://github.com/opentracing/opentracing.github.io/pull/19 done, I'm going to make appropriate changes to README's in `master` branches)

yurishkuro:
@bensigelman is there any known outstanding tasks here? If so, let's enumerate as checkboxes and link to issues in respective repos.

bensigelman:
@yurishkuro I think just https://github.com/opentracing/opentracing-python/pull/10 -- if you merge / LGTM that we are done.

yurishkuro:
Merged, and closing this.

PR #19, opened 2016-01-13
language-independent semantics documentation

bensigelman:
... and some other things.

Highly recommend clicking this button: 
image

Open to nits on presentation / organization... I would ask that concerns about the actual semantics themselves be opened as github issues proper, though.

Thanks!

### DIFF
--- a/implementors.md
+++ b/implementors.md
@@ -1,0 +1,13 @@
+---
+layout: page
+title: Implementing the OpenTracing APIs
+---
+
+OpenTracing makes tracing portable for those adding instrumentation to their production systems.
+
+That said, it also makes life easier for those implementing monitoring and tracing tools: rather than spending time and effort learning the idioms of numerous platforms, designing featureful tracing APIs, and maintaining them, projects which implement OpenTracing can focus on delivering value and making the moste of the instrumented data.
+
+Furthermore, in a codebase instrumented with OpenTracing, adding (or switching to) an OpenTracing-compatible tool is an `O(1)` operation for the programmer: it's really just a configuration change. Since instrumentation is often the greatest obstacle to adoption, OpenTracing support is a great way to put a moritoring and/or tracing tool in the hands of more developers quickly.
+
+For information about current reference implementations in particular platforms, please reach out to the OpenTracing authors via [email](https://groups.google.com/forum/#!forum/distributed-tracing) or
+[gitter](https://gitter.im/opentracing/public).
--- a/implementors.md
+++ b/implementors.md
@@ -5,9 +5,9 @@ title: Implementing the OpenTracing APIs
 
 OpenTracing makes tracing portable for those adding instrumentation to their production systems.
 
-That said, it also makes life easier for those implementing monitoring and tracing tools: rather than spending time and effort learning the idioms of numerous platforms, designing featureful tracing APIs, and maintaining them, projects which implement OpenTracing can focus on delivering value and making the moste of the instrumented data.
+That said, it also makes life easier for those implementing monitoring and tracing tools: rather than spending time and effort learning the idioms of numerous platforms, designing featureful tracing APIs, and maintaining them, projects which implement OpenTracing can focus on delivering value and making the most of the instrumented data.
 
-Furthermore, in a codebase instrumented with OpenTracing, adding (or switching to) an OpenTracing-compatible tool is an `O(1)` operation for the programmer: it's really just a configuration change. Since instrumentation is often the greatest obstacle to adoption, OpenTracing support is a great way to put a moritoring and/or tracing tool in the hands of more developers quickly.
+Furthermore, in a codebase instrumented with OpenTracing, adding (or switching to) an OpenTracing-compatible tool is an `O(1)` operation for the programmer: it's really just a configuration change. Since instrumentation is often the greatest obstacle to adoption, OpenTracing support is a great way to put a monitoring and/or tracing tool in the hands of more developers quickly.
 
 For information about current reference implementations in particular platforms, please reach out to the OpenTracing authors via [email](https://groups.google.com/forum/#!forum/distributed-tracing) or
 [gitter](https://gitter.im/opentracing/public).
--- a/spec.md
+++ b/spec.md
@@ -76,7 +76,7 @@ The `Tracer` interface must have the following capabilities:
 
 The `Span` interface must have the following capabilities:
 
-- Create and start a new child `Span` with a given operation name **(py: `start_child`, go: `StartChild`)**
+- Create and start a new child `Span` with a given operation name. The child `Span`'s `TraceContext` must descend from the parent `Span`'s `TraceContext`, and any trace attributes must propagate through into the child's `TraceContext`. **(py: `start_child`, go: `StartChild`)**
 - Finish the (already-started) `Span`.  Finish should be the last call made to any span instance, and to do otherwise leads to undefined behavior. **(py: `finish`, go: `Finish`)**
 - Set a key:value tag on the `Span`. The key must be a `string`, and the value must be either a `string`, a `boolean`, or a numeric type. Behavior for other value types is undefined. If multiple values are set to the same key (i.e., in multiple calls), implementation behavior is also undefined. **(py: `set_tag`, go: `SetTag`)**
 - Add a new info `Log` to the `Span`, accepting a message `string` and 0 or more payload arguments. The payload arguments may be of any type and arbitrary size, though implementations are not required to retain all payload arguments (or even all parts of all payload arguments). **(py: `info`, go: `Info`)**
--- a/spec.md
+++ b/spec.md
@@ -48,9 +48,9 @@ Every Span has zero or more **Logs**, each of which being a timestamped message 
 
 Every Span may also have zero or more key/value **Tags**, which do not have timestamps and simply annotate the spans.
 
-A **Trace Context** encapsulates the smallest amount of state needed to continue a trace across a process boundary (or, more generally, a serialization boundary). Put another way, a Span cannot be identified without a Trace Context, but a Trace Context can be (and is) serialized without a Span: Spans are identified by their Trace Context and add application-level instrumentation (logs, timing information, tags). Note that, in order to support *Trace Attributes* (see below) that propagate across a distributed trace transparently, those attributes must be represented by/within the Trace Context.
+Every Span is bound to a **Trace Context**. The Trace Context describes how the Span it's bound to fits in to the larger Trace. A Trace Context encapsulates the smallest amount of state needed to continue a Trace across a process boundary (or, more generally, a serialization boundary). Put another way, a Span cannot exist without a Trace Context, but a Trace Context can be (and is) serialized without a Span: Spans are built around their Trace Context and concern themselves with application-level instrumentation (logs, timing information, tags). Note that, in order to support *Trace Attributes* (see below) that propagate across a distributed Trace transparently, those attributes must be represented by/within the Trace Context.
 
-**Trace Attributes** are key/value pairs stored in a Trace Context and propagated _in-band_ to all future child Spans. Given a full-stack OpenTracing integration, Trace Attributes enable powerful functionality by transparently propagating arbitrary application data: for example, an end-user id may be added as a trace attribute in a mobile app, propagate (via the distributed tracing machinery) into the depths of a storage system, and recovered at the bottom of the stack to identify a particularly expensive SQL query.
+**Trace Attributes** are key/value pairs stored in a Trace Context and propagated _in-band_ to all future child Spans. Given a full-stack OpenTracing integration, Trace Attributes enable powerful functionality by transparently propagating arbitrary application data: for example, an end-user id may be added as a Trace Attribute in a mobile app, propagate (via the distributed tracing machinery) into the depths of a storage system, and recovered at the bottom of the stack to identify a particularly expensive SQL query.
 
 Trace Attributes come with powerful _costs_ as well; since the attributes are propagated in-band, if they are too large or too numerous they may reduce system throughput or contribute to RPC latency.
 
@@ -67,7 +67,7 @@ OpenTracing supports a number of different platforms, and of course the per-plat
 
 The `Tracer` interface must have the following capabilities:
 
-- Start a `Span` that has no parent **(py: `start_trace`, go: `StartTrace`)**
+- Start a `Span` that has no parent, commonly referred to as a *root* `Span` **(py: `start_trace`, go: `StartTrace`)**
 - Start a `Span` as a descendant of a parent `TraceContext` **(py: `join_trace`, go: `JoinTrace`)**
 - Start a `Span` explicitly built around a specific `TraceContext` **(py: `Span(trace_context)`, go: `StartSpanWithContext`)**
 
@@ -121,13 +121,15 @@ Implementations create `TraceContext` instances via a `TraceContextSource` inter
 - Create a new child `TraceContext` given a parent `TraceContext`. **(py: `new_child_trace_context`, go: `NewChildTraceContext`)**
 
 
-Additionally, a `TraceContextSource` must be able to encode and decode `TraceContext`s. In its encoded form, a `TraceContext` is separated into a *pair* of values: one represents the "span identity" – for example, in Zipkin or Dapper, this would be the `trace_id` and `span_id`; and the other encoded value represents the trace attributes. This separation of encoded state enables optimizations in certain binary protocols.
+Additionally, a `TraceContextSource` must be able to encode and decode `TraceContext`s. In its encoded form, a `TraceContext` is separated into a *pair* of values: one represents the "span identity" – for example, in Zipkin or Dapper, this would be the `trace_id` and `span_id`; and the other encoded value represents the trace attributes. This separation of encoded state enables optimizations in certain binary protocols. For instance, if a particular `TraceContext` has a fixed number of fixed-length "span identity" fields, then the span identity can be encoded into a fixed-length buffer (while of course the trace attributes cannot).
 
 Note that there is no expectation that different tracing systems encode a `TraceContext` in compatible ways. Though OpenTracing is agnostic about the tracing implementation, for successful inter-process handoff it's essential that the processes on both sides use the same tracing implementation.
 
 Depending on the language, support for `TraceContext` coding may involve extending or embedding `TraceContextEncoder` and `TraceContextDecoder` interfaces with the following capabilities:
 
+Encoding and decoding may involve either **"binary"** or **"text"** formats. The "binary" format is an implementation-specific byte array (as opposed to a unicode/utf8 string); it is opaque at the OpenTracing level. The "text" format is a platform-idiomatic map from (unicode) `string` to `string`.
+
 - Encoding a `TraceContext` as a pair of binary values **(py: `trace_context_to_binary`, go: `TraceContextToBinary`)**
-- Encoding a `TraceContext` as a pair of text-encoded `string->string` maps, where the keys are suitable for HTTP heders (see the notes about "trace attribute" keys above) **(py: `trace_context_to_text`, go: `TraceContextToText`)**
+- Encoding a `TraceContext` as a pair of text-encoded `string->string` maps, where the keys are suitable for HTTP headers (see the notes about "trace attribute" keys above) **(py: `trace_context_to_text`, go: `TraceContextToText`)**
 - Decoding a `TraceContext` given a pair of binary values **(py: `trace_context_from_binary`, go: `TraceContextFromBinary`)**
 - Decoding a `TraceContext` given a pair of textual `string->string` maps **(py: `trace_context_from_text`, go: `TraceContextFromText`)**
--- a/spec.md
+++ b/spec.md
@@ -3,13 +3,19 @@ layout: page
 title: Semantic Specification and Terminology
 ---
 
+## Terminology
+
 A **Trace** represents the potentially distributed, potentially concurrent data/execution path in a (potentially distributed, potentially concurrent) system.
 
-A **Span** represents a logical unit of work in the system that has a start time and an end time. Spans can be nested and ordered to model parent-child and casual relationships. A Trace can be thought of as a tree of Spans. Every Span has zero or more **Logs**, each of which being a timestamped message with an optional structured data payload of arbitrary size. Every Span may also have zero or more key/value **Tags**, which do not have timestamps and simply annotate the spans.
+A **Span** represents a logical unit of work in the system that has a start time and an end time. Spans may be nested and ordered to model parent-child and casual relationships. A Trace can be thought of as a tree of Spans.
+
+Every Span has zero or more **Logs**, each of which being a timestamped message with an optional structured data payload of arbitrary size.
+
+Every Span may also have zero or more key/value **Tags**, which do not have timestamps and simply annotate the spans.
 
 A **Trace Context** encapsulates the smallest amount of state needed to describe a Span's identity within a larger, potentially distributed trace, sufficient to propagate the context of a particular trace between processes.
 
-**Trace Attributes** are key/value pairs stored in a Trace Context and propagated _in-band_ to all future children Spans. Given a full-stack OpenTracing integration, Trace Attributes enable powerful functionality by transparently propagating arbitrary application data; e.g., from a mobile app all the way into the depths of a storage system. It comes with powerful _costs_ as well, since the attributes are propagated in-band, alongside the application data; use Trace Attributes with care.
+**Trace Attributes** are key/value pairs stored in a Trace Context and propagated _in-band_ to all future child Spans. Given a full-stack OpenTracing integration, Trace Attributes enable powerful functionality by transparently propagating arbitrary application data: e.g., from a mobile app all the way into the depths of a storage system. Trace attributes come with powerful _costs_ as well; since the attributes are propagated in-band, if they are too large they can reduce system throughput or contribute to RPC latency. Use Trace Attributes with care.
 
 **Trace Attributes** vs. **Span Tags**
 
--- a/spec.md
+++ b/spec.md
@@ -1,9 +1,9 @@
 ---
 layout: page
-title: Semantic Specification and Terminology
+title: Semantic Specification
 ---
 
-## Terminology
+## Concepts and Terminology
 
 A **Trace** represents the potentially distributed, potentially concurrent data/execution path in a (potentially distributed, potentially concurrent) system.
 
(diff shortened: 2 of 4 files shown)


implementors.md
```diff
@@ -0,0 +1,13 @@
+---
+layout: page
+title: Implementing the OpenTracing APIs
+---
+
+OpenTracing makes tracing portable for those adding instrumentation to their production systems.
+
+That said, it also makes life easier for those implementing monitoring and tracing tools: rather than spending time and effort learning the idioms of numerous platforms, designing featureful tracing APIs, and maintaining them, projects which implement OpenTracing can focus on delivering value and making the moste of the instrumented data.
```
slimsag: typo at the end:

> making the **moste** of the instrumented data.

spec.md
```diff
@@ -0,0 +1,81 @@
+---
+layout: page
+title: Semantic Specification
+---
+
+## Concepts and Terminology
+
+A **Trace** represents the potentially distributed, potentially concurrent data/execution path in a (potentially distributed, potentially concurrent) system.
```
slimsag: Would adding "It is a collection of one or more spans." at the end help clarify this a bit?

spec.md
```diff
@@ -0,0 +1,81 @@
...
+
+A **Span** represents a logical unit of work in the system that has a start time and an end time. Spans may be nested and ordered to model parent-child and casual relationships. A Trace can be thought of as a tree of Spans.
```
slimsag: Actually, re: my above comment.. just move `A Trace can be thought of as a tree of Spans.` into the `**Trace**` section instead

dankosaur: :+1:

spec.md
```diff
@@ -0,0 +1,81 @@
...
+
+Every Span has zero or more **Logs**, each of which being a timestamped message with an optional structured data payload of arbitrary size.
+
+Every Span may also have zero or more key/value **Tags**, which do not have timestamps and simply annotate the spans.
+
+A **Trace Context** encapsulates the smallest amount of state needed to describe a Span's identity within a larger, potentially distributed trace, sufficient to propagate the context of a particular trace between processes.
```
slimsag: I'd like to see more elaboration here as trace context is still a very abstract puzzle piece in my mind, so I think others would have trouble understanding from this description, too.

Is the trace context responsible for keeping track of the parent<->child span relationships? If so that'd be a good thing to explicitly mention here.

spec.md
```diff
@@ -0,0 +1,81 @@
...
+
+- Create and start a new child `Span` with a given operation name **(py: `start_child`, go: `StartChild`)**
+- Finish the (already-started) `Span`.  Finish should be the last call made to any span instance, and to do otherwise leads to undefined behavior. **(py: `finish`, go: `Finish`)**
+- Set a key:value tag on the `Span`. The key must be a `string`, and the value must be either a `string`, a `boolean`, or a numeric type. Behavior for other value types is undefined. If multiple values are set to the same key (i.e., in multiple calls), implementation behavior is also undefined. **(py: `set_tag`, go: `SetTag`)**
+- Add a new info `Log` to the `Span`, accepting a message `string` and 0 or more payload arguments. The payload arguments may be of any type and arbitrary size, though implementations are not required to retain all payload arguments (or even all parts of all payload arguments). **(py: `info`, go: `Info`)**
+- Add a new error `Log` to the `Span`; analagous to adding an info log aside from the severity. **(py: `error`, go: `Error`)**
+- Access the `TraceContext` associated with the `Span`. **(py: `trace_context`, go: `TraceContext()`)**
+
+
+#### TraceContext
```
dankosaur: I suspect that for most instrumentation, the encoders/decoders in the subsequent section are much more important than trace attributes.  Suggest those be pulled into a subsection here and mentioned first, with trace attributes in a second subsection below.  WDYT?

spec.md
```diff
@@ -0,0 +1,81 @@
...
+Every Span may also have zero or more key/value **Tags**, which do not have timestamps and simply annotate the spans.
+
+A **Trace Context** encapsulates the smallest amount of state needed to describe a Span's identity within a larger, potentially distributed trace, sufficient to propagate the context of a particular trace between processes.
+
+**Trace Attributes** are key/value pairs stored in a Trace Context and propagated _in-band_ to all future child Spans. Given a full-stack OpenTracing integration, Trace Attributes enable powerful functionality by transparently propagating arbitrary application data: e.g., from a mobile app all the way into the depths of a storage system. Trace attributes come with powerful _costs_ as well; since the attributes are propagated in-band, if they are too large they can reduce system throughput or contribute to RPC latency. Use Trace Attributes with care.
+
+**Trace Attributes** vs. **Span Tags**
+
+* Trace Attributes are propagated in-band across process boundaries. Span Tags are not propagated.
+* Span Tags are recorded off-band in the tracing system's storage. Trace Attributes are not recorded.
```
slimsag: s/off-band/out-of-band/ ?

It's silly of me to ask this, I know.. but could we define the in-band and off-band/out-of-band terminology here or link to a good definition? for those less familiar with the specifics, i.e. just using the API, I think they will be easily confused by this terminology (I was when it originally came up).

spec.md
```diff
@@ -0,0 +1,81 @@
...
+- Set a key:value tag on the `Span`. The key must be a `string`, and the value must be either a `string`, a `boolean`, or a numeric type. Behavior for other value types is undefined. If multiple values are set to the same key (i.e., in multiple calls), implementation behavior is also undefined. **(py: `set_tag`, go: `SetTag`)**
+- Add a new info `Log` to the `Span`, accepting a message `string` and 0 or more payload arguments. The payload arguments may be of any type and arbitrary size, though implementations are not required to retain all payload arguments (or even all parts of all payload arguments). **(py: `info`, go: `Info`)**
+- Add a new error `Log` to the `Span`; analagous to adding an info log aside from the severity. **(py: `error`, go: `Error`)**
+- Access the `TraceContext` associated with the `Span`. **(py: `trace_context`, go: `TraceContext()`)**
+
+
+#### TraceContext
+
+Every `TraceContext` must provide access to "trace attributes". A **trace attribute** is a key:value pair associated with a TraceContext **that also propagates to future TraceContext children** per TraceContext.NewChild, and that propagates across process boundaries in-band with application data.
```
slimsag: s/a trace attributes is a/trace attributes are a/ ?

spec.md
```diff
@@ -0,0 +1,81 @@
...
+
+Trace attributes are powerful, especially given an OpenTracing integration that extends across many layers in a distributed stack. For example, arbitrary application data from a mobile app can travel, transparently, all the way into the depths of a storage system). These powerful capabilities come with some powerful costs: use this feature with care.
```
slimsag: This seems to be mostly (completely?) covered by the Trace Attributes section on line 18.

spec.md
```diff
@@ -0,0 +1,81 @@
...
+
+Also, trace attribute keys have a restricted format: implementations may wish to use them as HTTP header keys (or key suffixes), and of course HTTP headers are case insensitive.
+
+As such, trace attribute keys MUST match the regular expression `(?i:[a-z0-9][-a-z0-9]*)`, and – per the `?i:` – they are case-insensitive.  That is, the trace attribute key must start with a letter or number, and the remaining characters must be letters, numbers, or hyphens. See CanonicalizeTraceAttributeKey(). If a trace attribute key does not meet these criteria, SetTraceAttribute() results in undefined behavior.
```
slimsag: Since it's a spec / language independent, we shouldn't reference an implementation `CanonicalizeTraceAttributeKey` here. I think it'd be fine to just remove that part.

spec.md
```diff
@@ -0,0 +1,81 @@
...
+
+In any case, the programmatic `TraceContext` interface is deceptively simple:
+
+- Set a trace attribute, which is a simple string:string pair. Note that newly-set trace attributes are only guaranteed to propagate to *future* children of the `TraceContext` and/or its associated `Span`. **(py: `set_trace_attribute`, go: `SetTraceAttribute`)**
```
slimsag: The `it's` part of `the `TraceContext` and/or its associated` has me a bit confused. Are we saying that trace attributes:

1. "are only guaranteed to propagate to future children of the TraceContext and/or (future child trace context) spans"?
1. "are only guaranteed to propagate to future children of the TraceContext and/or (current/present trace context) spans"?

?

spec.md
```diff
@@ -0,0 +1,81 @@
...
+
+#### TraceContextSource, TraceContextEncoder, TraceContextDecoder
+
+Implementations create `TraceContext` instances via a `TraceContextSource` interface; it must have the following capabilities:
+
+- Create a new parent-less ("root") `TraceContext`. **(py: `new_root_trace_context`, go: `NewRootTraceContext`)**
+- Create a new child `TraceContext` given a parent `TraceContext`. **(py: `new_child_trace_context`, go: `NewChildTraceContext`)**
+
+
+Additionally, a `TraceContextSource` must be able to encode and decode `TraceContext`s. In its encoded form, a `TraceContext` is separated into a *pair* of values: one represents the "span identity" – for example, in Zipkin or Dapper, this would be the `trace_id` and `span_id`; and the other encoded value represents the trace attributes. This separation of encoded state enables optimizations in certain binary protocols.
```
slimsag: Should make it clear here that the encoding is not standardized anywhere, i.e. not required to be compatible with other tracing systems.

slimsag:
This is amazing! Thanks a ton @bensigelman ! I left a bunch of comments, but aside from those, LGTM

spec.md
yurishkuro: Agree that it needs a bit more, although @slimsag - no, it's not responsible for parent-child relationships, not explicitly anyway. E.g. if you're familiar with Zipkin's (trace_id, span_id, parent_span_id) tuple, you may notice that  parent_span_id did not have to be propagated in-band across RPCs, it could've been reported off-band.

@bensigelman the current wording makes emphasis on TC being the _identity_ of the span, maybe we can rephrase  that it actually has two roles, describing the span identity, and a container for distributed context.

spec.md
```diff
@@ -0,0 +1,81 @@
...
+
+Every Span has zero or more **Logs**, each of which being a timestamped message with an optional structured data payload of arbitrary size.
+
+Every Span may also have zero or more key/value **Tags**, which do not have timestamps and simply annotate the spans.
+
+A **Trace Context** encapsulates the smallest amount of state needed to describe a Span's identity within a larger, potentially distributed trace, sufficient to propagate the context of a particular trace between processes.
+
+**Trace Attributes** are key/value pairs stored in a Trace Context and propagated _in-band_ to all future child Spans. Given a full-stack OpenTracing integration, Trace Attributes enable powerful functionality by transparently propagating arbitrary application data: e.g., from a mobile app all the way into the depths of a storage system. Trace attributes come with powerful _costs_ as well; since the attributes are propagated in-band, if they are too large they can reduce system throughput or contribute to RPC latency. Use Trace Attributes with care.
```
yurishkuro: capitalize Attributes in the sentence "TAs come with powerful costs as well"

spec.md
```diff
@@ -0,0 +1,81 @@
...
+
+OpenTracing supports a number of different platforms, and of course the per-platform APIs try to blend in to their surroundings and do as the Romans do. That said, each platform API must model a common set of semantics for the core tracing concepts described above. In this document we attempt to describe those concepts and semantics in a language- and platform-agnostic fashion.
+
+#### Tracer
+
+The `Tracer` interface must have the following capabilities:
+
+- Start a `Span` that has no parent **(py: `start_trace`, go: `StartTrace`)**
+- Start a `Span` given a parent `TraceContext` **(py: `join_trace`, go: `JoinTrace`)**
+- Start a `Span` associated with a specific `TraceContext` **(py: `start_span_with_context`, go: `StartSpanWithContext`)**
```
yurishkuro: I would lose "associated", it sounds ambiguous. Maybe s/associated with/identified by/

spec.md
```diff
@@ -0,0 +1,81 @@
...
+
+
+#### Span
+
+The `Span` interface must have the following capabilities:
+
+- Create and start a new child `Span` with a given operation name **(py: `start_child`, go: `StartChild`)**
+- Finish the (already-started) `Span`.  Finish should be the last call made to any span instance, and to do otherwise leads to undefined behavior. **(py: `finish`, go: `Finish`)**
+- Set a key:value tag on the `Span`. The key must be a `string`, and the value must be either a `string`, a `boolean`, or a numeric type. Behavior for other value types is undefined. If multiple values are set to the same key (i.e., in multiple calls), implementation behavior is also undefined. **(py: `set_tag`, go: `SetTag`)**
+- Add a new info `Log` to the `Span`, accepting a message `string` and 0 or more payload arguments. The payload arguments may be of any type and arbitrary size, though implementations are not required to retain all payload arguments (or even all parts of all payload arguments). **(py: `info`, go: `Info`)**
```
yurishkuro: Clarification - are the implementations required to retain the `message` string?

Maybe worth repeating that it's the Span's responsibility to capture the timestamp of the log.

spec.md
yurishkuro: Hm, I would disagree. Trace Attributes are the core feature required for distributed context propagation, so it's best to explain it first, before diving into how those attributes are serialized.

spec.md
yurishkuro: s/must provide access/must provide read/write access/

yurishkuro: I think we moved `TraceContext.NewChild` to TraceContextSource.

spec.md
yurishkuro: agreed

spec.md
```diff
@@ -0,0 +1,81 @@
...
+
+
+#### TraceContext
+
+Every `TraceContext` must provide access to "trace attributes". A **trace attribute** is a key:value pair associated with a TraceContext **that also propagates to future TraceContext children** per TraceContext.NewChild, and that propagates across process boundaries in-band with application data.
+
+Trace attributes are powerful, especially given an OpenTracing integration that extends across many layers in a distributed stack. For example, arbitrary application data from a mobile app can travel, transparently, all the way into the depths of a storage system). These powerful capabilities come with some powerful costs: use this feature with care.
+
+Also, trace attribute keys have a restricted format: implementations may wish to use them as HTTP header keys (or key suffixes), and of course HTTP headers are case insensitive.
+
```
yurishkuro: I'd keep it as one paragraph.

spec.md
yurishkuro: I suggest s/children/descendants/

spec.md
yurishkuro: Actually there are other places where we're in need of such verb indicating relationship between Span and TC. Maybe we should define that verb in the terminology section. Still not sure "associated" is the right verb though, the relationship is stronger than a mere association.

spec.md
```diff
@@ -0,0 +1,81 @@
...
+As such, trace attribute keys MUST match the regular expression `(?i:[a-z0-9][-a-z0-9]*)`, and – per the `?i:` – they are case-insensitive.  That is, the trace attribute key must start with a letter or number, and the remaining characters must be letters, numbers, or hyphens. See CanonicalizeTraceAttributeKey(). If a trace attribute key does not meet these criteria, SetTraceAttribute() results in undefined behavior.
+
+In any case, the programmatic `TraceContext` interface is deceptively simple:
+
+- Set a trace attribute, which is a simple string:string pair. Note that newly-set trace attributes are only guaranteed to propagate to *future* children of the `TraceContext` and/or its associated `Span`. **(py: `set_trace_attribute`, go: `SetTraceAttribute`)**
+- Get a trace attribute by key. **(py: `get_trace_attribute`, go: `TraceAttribute`)**
+
+
+#### TraceContextSource, TraceContextEncoder, TraceContextDecoder
+
```
yurishkuro: maybe worth mentioning somewhere that a tracer implements all three interfaces and developers using the API to instrument their applications only interact with these interfaces via the tracer object.

yurishkuro:
Looks great!

spec.md
dankosaur: @yurishkuro I am not sure I follow, so this might be a good sanity-check as to clarity of this documentation section to someone who's (still) a bit of an outsider on this project then.

From reading the code, I thought that the IDs are built in to TraceContext, and that Attributes are optional extensions to this.  Under that assumption, the main use-case instrumenters will have for TraceContext is using it to propagate the associated IDs.  Optionally, the Attributes can be used to augment this with information that might be useful at an instrumentation point later in the trace process.

On the other hand, from this documentation, I got the gist that Annotations are the biggest part of TraceContext and it wasn't clear that TraceContext is responsible for ID propagation.  I also don't come away entirely clear what the most common use-cases for Attributes would be.  I remember Rodrigo's discussion of "baggage" influencing sampling or even execution later in the trace chain, but I'm not aware of what use-cases are currently in tracers out in the wild--some examples might help?

(It's possible that this spec is ahead of implementations like zipkin-like and we are planning to use Attributes to capture trace/span IDs--if so, that explains my confusion!)

yurishkuro: @dankosaur You're right, Dan, attributes are the equivalent of "baggage" (let me use this term for disambiguation), and they are completely optional and not required for trace/span ID. But they are essential for "baggage" propagation if the developers have a need for that. At my company a lot of people are actually very excited about "baggage", planning to use that for security credentials propagation, and for resource usage cost accounting.

That's why my preference is to group/separate functionality discussion into
- propagation functionality
    - trace/span ID
    - baggage
- serialization functionality

NB: we had a side discussion elsewhere that propagation as a topic might be more complex, not just "to all descendants", but "direct children only", or even "ancestors".

implementors.md
```diff
@@ -0,0 +1,13 @@
...
+
+Furthermore, in a codebase instrumented with OpenTracing, adding (or switching to) an OpenTracing-compatible tool is an `O(1)` operation for the programmer: it's really just a configuration change. Since instrumentation is often the greatest obstacle to adoption, OpenTracing support is a great way to put a moritoring and/or tracing tool in the hands of more developers quickly.
```
adriancole: there are some subtleties here that I feel very comfortable with. For example, it suggests use cases without implying that OpenTracing tools themselves have compatible features.

spec.md
adriancole: kindof like duration as opposed to end time, as there's existing code that can do a better job by supplying explicit duration. That said I think that if this is just to introduce things conceptually, it is fair to leave weeds in the weeds.

spec.md
adriancole: If folks feel generally useful, I'd like an issue that talks about parents a little better. Ex in appneta, it seems addEdge is a user api, which would map to a tag. In zipkin parent is sent out-of-band, but it is still not a user-level (tag) concern. Not saying we need to have an opinion per-se, but this topic is bound to be a bit confusing as parents are treated at different levels of abstraction.

spec.md
adriancole: I think this description is quite good.

spec.md
adriancole: Trace attributes could be recorded. Ex. in zipkin, the propagated ids are recorded. Also, the DEBUG propagation flag is recorded into the span (so that it can be processed by an out-of-band collection)

spec.md
adriancole: @karlunho I recall you have a use case, from AWS lambda? Basically, seems like you are resuming state from an externalized context. (sorry about the terrible sentence)

spec.md
adriancole: also, in zipkin this is "annotation" except presumed to be a string. While mostly used for single-word, timestamped tags, like "sr", it is also used as a log. For example, I've noticed people dumping GC objects into this, and sometimes printing things like "finagle.retried".

spec.md
```diff
@@ -0,0 +1,81 @@
...
+#### Span
+
+The `Span` interface must have the following capabilities:
+
+- Create and start a new child `Span` with a given operation name **(py: `start_child`, go: `StartChild`)**
+- Finish the (already-started) `Span`.  Finish should be the last call made to any span instance, and to do otherwise leads to undefined behavior. **(py: `finish`, go: `Finish`)**
+- Set a key:value tag on the `Span`. The key must be a `string`, and the value must be either a `string`, a `boolean`, or a numeric type. Behavior for other value types is undefined. If multiple values are set to the same key (i.e., in multiple calls), implementation behavior is also undefined. **(py: `set_tag`, go: `SetTag`)**
+- Add a new info `Log` to the `Span`, accepting a message `string` and 0 or more payload arguments. The payload arguments may be of any type and arbitrary size, though implementations are not required to retain all payload arguments (or even all parts of all payload arguments). **(py: `info`, go: `Info`)**
+- Add a new error `Log` to the `Span`; analagous to adding an info log aside from the severity. **(py: `error`, go: `Error`)**
```
adriancole: interesting feature

spec.md
adriancole: exported mental note: got this far.. will resume later (and delete this)

spec.md
karlunho: yeap.

spec.md
dankosaur: @yurishkuro I like that bullet outline above.  I think some combination of this and the clarification @slimsag requested in the summary section above will address my concerns.

Propagation policy description may be aided by some ascii art or other diagram.

spec.md
bensigelman: makes sense... I will reword.

spec.md
bensigelman: yes, "out of band" is the wording I would shoot for. I think this block was moved from an existing doc or something. I'll clarify the semantics regardless.

spec.md
bensigelman: Agreed about "associated" being suboptimal: maybe "identified by"?

spec.md
bensigelman: frankly I'm a little antsy about the "message"... namely, should OpenTracing concern itself with the way the format string works? (See the description in the golang API: https://github.com/opentracing/api-go/blob/master/opentracing/span.go#L26, for example)

But I will clarify per @yurishkuro request.

spec.md
bensigelman: (important for sampling systems and UIs to treat error spans specially)

spec.md
bensigelman: (I will try to do something reasonable tomorrow. @dankosaur I `<3` ASCII art and will try to come up with something on that front, too :)

spec.md
bensigelman: @yurishkuro very true re `NewChild`, good catch!

spec.md
bensigelman: I will make a diagram for this... pictures, 1000s of words, etc, etc.

spec.md
bensigelman: good idea.

bensigelman:
All: thanks for the great feedback. I have had an insane couple of days but have literally blocked of my calendar for an hour tomorrow AM to respond to the great suggestions. If I didn't reply to a comment, that means I'm going to make the requested change (not that I'm going to ignore it!).

Thanks again.

spec.md
yurishkuro: Considering that Zipkin is an important use case given its popularity, it's a bit sad that we don't capture the _events_ (aka "annotations" in Zipkin). We sort-of do with `info(message)`, but there is a semantic difference between that and `event(event_name)`, because `event_name` would be considered a more concise string (not a sentence / free form text). For indexing Zipkin today supports searches `where event_name="..."`, which makes much more sense than a search `where message="..."`.

I am not sure how best to express it via API, because always requiring event_name in the `info()` method is also weird.   However we do it, the main semantic difference would be that the tracer SHOULD (or even MUST) retain the `event_name`, while it only MIGHT retain the `message` and `payload`

bensigelman: one option (which would definitely be a diff on the current API):
- introduce a new `LogItem` or `LogLine` (or just `Log`) interface
- all of the following would return a `LogItem` (or whatever we call it):
    - have Info/Error just adopt platform conventions around log message formatting (would be `Infof`/`Errorf` in golang, e.g.)
    - add an independent `Payload(object)` method that takes arbitrary data structures
    - add an independent `Event(string)` method that's similar to what Zipkin does now
- the `LogItem` interface would also have analogous Info/Error/Payload/Event methods which could be chained

Calling code would then look like (in Go, e.g.):

```
    span.Infof("A message about an int %d and something else: %v", anInt, whatever)
    span.Event("something_stable").Payload(bigObject)
```

etc.

Obviously there are lots of variants on the above. Thoughts??

yurishkuro: That's assuming `Event()` method needs a payload. Maybe if it didn't (since payload can be logged with `Info`), then the only real change is adding an extra `Event` method that returns the span (which reminded me: https://github.com/opentracing/api-python/issues/7)

spec.md
```diff
@@ -0,0 +1,81 @@
...
+- Start a `Span` that has no parent **(py: `start_trace`, go: `StartTrace`)**
+- Start a `Span` given a parent `TraceContext` **(py: `join_trace`, go: `JoinTrace`)**
+- Start a `Span` associated with a specific `TraceContext` **(py: `start_span_with_context`, go: `StartSpanWithContext`)**
+
+
+#### Span
+
+The `Span` interface must have the following capabilities:
+
+- Create and start a new child `Span` with a given operation name **(py: `start_child`, go: `StartChild`)**
```
bcronin: Two cents: it might be nice to explicitly spell out here what *must* happen and what is implementation dependent with regards to the inheritance from parent to child of both trace attributes and tags.

spec.md
bensigelman: I'm lukewarm on `Event` -- so the motivation is to give the tracing impl a hint as to whether something (the message xor event) should be indexed? Anyway, if you want to make an issue out of it, sgtm, but I will try not to discuss too much more on this thread since I wanted to stick to comments about the wording rather than semantics. (My fault for adding the proposal two messages ago!)

spec.md
bensigelman: @dankosaur let me know if the ASCII thing I added here helps.

bensigelman:
@slimsag @yurishkuro @adriancole @dankosaur @bcronin 

Thank you all for the comments! I just pushed an iteration that attempts to address most of them. I did *not* attempt to rejigger the inheritance stuff that we've been discussing elsewhere, nor did I settle the log message thing that @yurishkuro and I were discussing.

Proposal: I fix any lingering issues and merge this; we then work through the blocking "structural" API issues (the inheritance thing, consolidating encoding, etc), and I update this doc accordingly (either before or in parallel with the actual platform-specific code changes, either way).

spec.md
adriancole: On events and payload. I agree payload isn't required for event. Ex maybe
we can contain the arbitrary payload parts tightly somewhere perhaps only
in the logging api.

I would go so far as to say id like to avoid adding more hooks that imply
tracing systems need/or are also a blobstore.

spec.md
```diff
@@ -0,0 +1,133 @@
...
+
+## Platform-Independent API Semantics
+
+OpenTracing supports a number of different platforms, and of course the per-platform APIs try to blend in to their surroundings and do as the Romans do. That said, each platform API must model a common set of semantics for the core tracing concepts described above. In this document we attempt to describe those concepts and semantics in a language- and platform-agnostic fashion.
+
+#### Tracer
+
+The `Tracer` interface must have the following capabilities:
+
+- Start a `Span` that has no parent **(py: `start_trace`, go: `StartTrace`)**
```
yurishkuro: add ", often referred to as *root* span" ?

spec.md
```diff
@@ -0,0 +1,133 @@
...
+
+Note that there is no expectation that different tracing systems encode a `TraceContext` in compatible ways. Though OpenTracing is agnostic about the tracing implementation, for successful inter-process handoff it's essential that the processes on both sides use the same tracing implementation.
+
+Depending on the language, support for `TraceContext` coding may involve extending or embedding `TraceContextEncoder` and `TraceContextDecoder` interfaces with the following capabilities:
+
+- Encoding a `TraceContext` as a pair of binary values **(py: `trace_context_to_binary`, go: `TraceContextToBinary`)**
+- Encoding a `TraceContext` as a pair of text-encoded `string->string` maps, where the keys are suitable for HTTP heders (see the notes about "trace attribute" keys above) **(py: `trace_context_to_text`, go: `TraceContextToText`)**
```
yurishkuro: typo: heders

spec.md
```diff
@@ -0,0 +1,133 @@
...
+```
+
+A **Trace** represents the potentially distributed, potentially concurrent data/execution path in a (potentially distributed, potentially concurrent) system. A Trace can be thought of as a tree of Spans. (See the ASCII diagrams above)
+
+A **Span** represents a logical unit of work in the system that has a start time and a duration. Spans may be nested and ordered to model parent-child and casual relationships. Each span has an **operation name**, a presumably human-readable string which concisely names the work done by the span (e.g., an RPC method name, a function name, or the name of a subtask within a larger computation).
+
+Every Span has zero or more **Logs**, each of which being a timestamped message with an optional structured data payload of arbitrary size.
+
+Every Span may also have zero or more key/value **Tags**, which do not have timestamps and simply annotate the spans.
+
```
yurishkuro: Maybe we can add: "Every Span has a unique identity determined by the *Trace Context*".

yurishkuro:
> Proposal: I fix any lingering issues and merge this;

+1

bensigelman:
I made adjustments based on @yurishkuro's most recent round of suggestions and added more detail about the codec stuff per @bcronin's comment off of this PR.

I'm going to merge this now so I can make progress on https://github.com/opentracing/opentracing.github.io/issues/10. That said, if there are concerns with anything in this doc – big or small – please bring them up and I'll open a new PR accordingly.

dankosaur:
@bensigelman :+1: a merge is good to clean up the discussion thread.

PR #34, opened 2016-01-22
Add a page illustrating common instrumentation use cases

yurishkuro:


use-cases.md
```diff
@@ -0,0 +1,143 @@
...
+
+This page aims to illustrate the common use cases that developers who instrument their applications with OpenTracing API need to deal with.
+
+### Tracing Server Endpoints
+
+When a server receives a request and needs to start tracing, it generally needs to go through three steps:
+
+  1. Attempt to deserialize Trace Context from the incoming request, in case the trace has already been started by the client.
+  1. Depending on the outcome, either start a new trace, or join the existing trace using deserialized Trace Context. This operation starts a new Span.
+  1. Store the newly created span in some _request context_ that is propagated throughout the application, either by application code, or by the RPC framework.
```
dankosaur: Might be worth adding "4. Finally, close the trace using `span.finish()` when the server has finished processing the request"

yurishkuro: @dankosaur good catch, updated.

use-cases.md
```diff
@@ -0,0 +1,143 @@
...
+Let's assume that we have an HTTP server, and the Trace Context is propagated from the client via HTTP headers, accessible via `request.headers`:
+
+```python
+    context = tracer.trace_context_from_text(
+        trace_context_id=request.headers,
+        trace_attributes=request.headers
+    )
+```
+
+Here we set both arguments of the decoding method to the `headers` map. The Tracer object will have the knowledge which headers it needs to read in order to reconstruct Trace Context, which is comprised of a trace context ID and trace attributes.
```
dankosaur: I was reminded that with custom attributes we have the potential for naming collision with common HTTP headers -- do you think this is worth adding a note about in spec.md?

use-cases.md
```diff
@@ -0,0 +1,143 @@
...
+
+```python
+    if context is None:
+        span = tracer.start_trace(operation_name=operation)
+    else:
+        span = tracer.join_trace(operation_name=operation,
+                                 parent_trace_context=context)
+    span.set_tag('http.method', request.method)
+    span.set_tag('http.url', request.full_url)
+```
```
dankosaur: Should we add a convenience `start_or_join_trace()` method that makes this top part a one-liner?

dankosaur:
This is very clear and helpful @yurishkuro

use-cases.md
yurishkuro: the way it's been implemented in a few places so far is that the custom attributes are prefixed by some string specific to the tracer. For example, our internal tracer serializes id into Uber-Trace-Id and attributes into `UberCtx-***`. 

But it's definitely worth making explicit somewhere. I would prefer it to be in the docstring of the actual interface classes that people will implement, rather than in semantic documentation.

use-cases.md
```diff
@@ -0,0 +1,144 @@
...
+    # client side
+    span.trace_context.set_attribute('auth-token', '.....')
+
+    # server side (one or more levels down from the client)
+    token = span.trace_context.get_attribute('auth-token')
+```
+
+### Logging Events
+
+We have already used `log_event_with_payload` in the client span use case. Events can be logged without a payload, and not just where the span is being created / finished. For example, the application may record a cache miss event in the middle of execution, as long as it can get access to the current span from request context:
```
dankosaur: add ~ "These events also have a timestamp associated with them, which distinguishes them from a tag applying to the entire span." ?

use-cases.md
```diff
@@ -0,0 +1,144 @@
...
+        for key, value in h_ctx.iteritems():
+            request.add_header(key, value)
+        if h_attr:
+            for key, value in h_attr.iteritems():
+                request.add_header(key, value)
+
+        # define a callback where we can finish the span 
+        def on_done(future):
+            if future.exception():
+                span.log_event_with_payload('exception', exception)
```
dankosaur: I know there is still some debate on the `Error` event spec in https://github.com/opentracing/opentracing-go/pull/38 , but if we do end up with that included, this is a great place to show it off.

use-cases.md
yurishkuro: done, and I also added a whole use case (unsupported atm) about recording externally captured timestamps.

public/js/toc.js
```diff
@@ -0,0 +1,89 @@
+// https://github.com/ghiculescu/jekyll-table-of-contents
+(function($){
```
bensigelman: why doesn't this use `opentracing-javascript`?! :)

use-cases.md
```diff
@@ -0,0 +1,150 @@
+---
+layout: page
+title: Instrumentation for Common Use Cases
+---
+<div id="toc"></div>
+
+This page aims to illustrate the common use cases that developers who instrument their applications with OpenTracing API need to deal with.
```
bensigelman: nit: s/the common/common/, s/their applications/their applications and libraries/ ?

use-cases.md
```diff
@@ -0,0 +1,150 @@
...
+  1. Store the newly created span in some _request context_ that is propagated throughout the application, either by application code, or by the RPC framework.
+  1. Finally, close the span using `span.finish()` when the server has finished processing the request.
+
+#### Deserializing Trace Conext from Incoming Request
+
+Let's assume that we have an HTTP server, and the Trace Context is propagated from the client via HTTP headers, accessible via `request.headers`:
+
+```python
+    context = tracer.trace_context_from_text(
+        trace_context_id=request.headers,
```
bensigelman: maybe add a comment that the maps passed to `trace_context_from_text` can be a superset of the fields that are required (hence passing the same headers in twice).

bensigelman: oh, just saw the note below. disregard!

use-cases.md
bensigelman: nit: "The Tracer object will have the knowledge which headers it needs to read ..." -> "The Tracer object knows which headers it needs to read ..."

bensigelman: micronit / common English bug: when people write "is comprised of", they almost always mean "comprises". So, "... Trace Context, which comprises a trace context ID and trace attributes."

http://public.wsu.edu/~brians/errors/comprised.html

use-cases.md
bensigelman: 1

bensigelman:
I decided midstream that nitpicking at a minute level was probably annoying and certainly not a good use of your time! So I will stop.

The only substantial request I'd have is that we include – probably at the top – a totally utterly vanilla span that just starts with an operation name and stops within the same function scope. And maybe a second example where we create a child span in-process, also totally vanilla.

yurishkuro:
I don't mind nit-picking, but it will probably be less work for you, after I merge this, to do another commit by simply changing the wording.

I will add a vanilla span example, but I didn't what were the _two_ things you mentioned, how are they different?

Vanilla example:

```python
    def my_function():
        span = get_current_span().start_child('my-function') if get_current_span() else None
        try:
            . . . # business logic
        finally:
            if span:
                span.finish()
```

use-cases.md
yurishkuro: :heart: the reference :-)

bensigelman:
I guess I was imagining – just for the sake of starting simple – something like

```
def my_top_level_function():
    trivial_span = tracer.start_trace("top-level-function")
    try:
        . . .
    finally:
        trivial_span.finish()
```

Then of course things can get more complex from there.

yurishkuro:
@bensigelman comments addressed

bensigelman:
great, LGTM, thanks! And agreed, if anyone cares I can take a pass at make my Composition professor happy with some silly line-edits as a separate PR. (Or not :)

use-cases.md
michaelsembwever: I agree, this was raised in https://github.com/opentracing/opentracing-java/pull/7/files#r50501336

Having the following API would be more intuitive and less intrusive instrumentation code

    Span Tracer.startTrace(String operation, @Nullable SpanContext cxt);

Whether the method is called startTrace(..) or trace(..) or createSpan(..) is more on the bikeshed side of things. My vote would be on createSpan(..) since that's what it does, implicit in it returns a Span object. Whether a trace is started or joined to is simply a consequence to the action.

use-cases.md
michaelsembwever: I'm in favour of 
 
    span.event(string, optional<payload>)

It's a nice clean simple API, that supports the different use-cases we've seen.

use-cases.md
```diff
@@ -0,0 +1,184 @@
...
+```python
+    def traced_request(request, operation, http_client):
+        # retrieve current span from propagated request context
+        parent_span = get_current_span()
+
+        # start a new child span or a brand new trace if no parent
+        if parent_span is None:
+            span = tracer.start_trace(operation_name=operation)
+        else:
+            span = parent_span.start_child(operation_name=operation)
```
michaelsembwever: I believe the API can do this as a one-liner.
The parent_span also has a SpanContext, using that we can use a simplified method along the lines of

    Span Tracer.startTrace(String operation, @Nullable SpanContext cxt);

use-cases.md
```diff
@@ -0,0 +1,184 @@
...
+The client and server examples above took care of propagating Trace Context on the wire, which includes any Trace Attributes.  The client may use the Trace Attributes to pass additional data to the server and any other downstream server it might call.
+
+```python
+
+    # client side
+    span.trace_context.set_attribute('auth-token', '.....')
+
+    # server side (one or more levels down from the client)
+    token = span.trace_context.get_attribute('auth-token')
+```
```
michaelsembwever: This does scare me. The OpenTracing API is going to get abused on this front.

The question i was asking in https://github.com/opentracing/opentracing-java/pull/7#discussion_r50502226

Another example is taking the correlation id out of the span_context (what's still called trace_context here) and putting it into the logging context (what log4j-1 calls the MDC and what log4j-2 calls the thread context) so that systems like Zipkin and Kibana can be used in combination.

I would rather see the logging example used here instead of "auth-token". The logging example keeps the suggestion of how to use the context propagation at the infrustructure/systems level, rather than it-can-be-used-for-whatever-you-like in your application code.

use-cases.md
michaelsembwever: where's the example of recording externally captured timestamps?

use-cases.md
```diff
@@ -0,0 +1,184 @@
...
+### Tracing Server Endpoints
+
+When a server wants to trace execution of a request, it generally needs to go through these steps:
+
+  1. Attempt to deserialize Trace Context from the incoming request, in case the trace has already been started by the client.
+  1. Depending on the outcome, either start a new trace, or join the existing trace using deserialized Trace Context. This operation starts a new Span.
+  1. Store the newly created span in some _request context_ that is propagated throughout the application, either by application code, or by the RPC framework.
+  1. Finally, close the span using `span.finish()` when the server has finished processing the request.
+
+#### Deserializing Trace Conext from Incoming Request
```
michaelsembwever: s/Conext/Context/

use-cases.md
```diff
@@ -0,0 +1,184 @@
...
+
+Let's assume that we have an HTTP server, and the Trace Context is propagated from the client via HTTP headers, accessible via `request.headers`:
+
+```python
+    context = tracer.trace_context_from_text(
+        trace_context_id=request.headers,
+        trace_attributes=request.headers
+    )
```
michaelsembwever: The problem I have with this API design is really nothing more than the use of the words serialisation and deserialisation.

While serialisation is obviously a consequence to this behind the scenes, it's not actually what the user cares about here, and therefore it's not intuitive API.

What this is really about is the transforming the state from the protocol's object to the tracing's context.

And I would say that the serialisation is the writing of bytes or characters behind the scenes into streams/buffers.

Maybe we should be looking at how other frameworks and SPI handle such transformer objects.

adriancole: seems this doc is now out of date or soon will be due to https://github.com/opentracing/opentracing-go/pull/40

use-cases.md
```diff
@@ -0,0 +1,184 @@
...
+    if context is None:
+        span = tracer.start_trace(operation_name=operation)
+    else:
+        span = tracer.join_trace(operation_name=operation,
+                                 parent_trace_context=context)
+    span.set_tag('http.method', request.method)
+    span.set_tag('http.url', request.full_url)
+```
+
+The `operation` above refers to the name the server wants to give to the Span. For example, if the HTTP request was a POST against `/save_user/123`, the operation name can be set to `post:/save_user/`. OpenTracing API does not dictate how application calls the spans.
```
adriancole: good note.

use-cases.md
```diff
@@ -0,0 +1,184 @@
...
+
+The `set_tag` calls are examples of recording additional information in the span about the request.
+
+#### In-Process Request Context Propagation
+
+Request context propagation refers to application's ability to associate a certain _context_ with the incoming request such that this context is accessible in all other layers of the application within the same process. It can be used to provide application layers with access to request-specific values such as the identity of the end user, authorization tokens, and the request's deadline. It can also be used for transporting the current tracing Span.
+
+The methods of request context propagation are outside the scope of the OpenTracing API, but it is worth mentioning them here to better understand the following sections. There are two commonly used methods of propagation:
```
adriancole: supernit: `The methods of request context propagation`. we literally do provide methods to propagate request data. `Implementation of request context propagation` unloads that sentence

use-cases.md
```diff
@@ -0,0 +1,184 @@
...
+
+  1. Explicit propagation - the application code is structured to pass around a certain context object. For example, see https://blog.golang.org/context.
```
adriancole: nit: For example (code snippet) then link. that makes lazy readers understand more

use-cases.md
```diff
@@ -0,0 +1,184 @@
...
+  1. Implicit propagation - the context is stored in platform-specific storage that allows it to be retrieved from any place in the application. Often used by RPC frameworks by utilizing such mechanisms as thread-local or continuation-local storage, or even global variables (in case of single-threaded processes).
```
adriancole: The reader of this doc could be an API or an SPI person. For example, the latter will likely build implicit with an explicit context. Not sure the best way to say this, but this made me think we shouldn't imply these are different solutions. 

Explicit or Implicit here is mainly how you get a handle on a context, ex via passing around, or via Dependency injection, or static. Maybe library vs framework is really the heart of this topic.

use-cases.md
adriancole: +1

use-cases.md
```diff
@@ -0,0 +1,184 @@
...
+
+        # start a new child span or a brand new trace if no parent
+        if parent_span is None:
+            span = tracer.start_trace(operation_name=operation)
+        else:
+            span = parent_span.start_child(operation_name=operation)
+        span.set_tag('server.http.url', request.full_url)
+
+        # encode Trace Context into HTTP request headers
+        h_ctx, h_attr = tracer.trace_context_to_text(
```
adriancole: this section will need complete rewrite I think, but I get where it is at.

use-cases.md
```diff
@@ -0,0 +1,184 @@
...
+    def traced_request(request, operation, http_client):
+        # retrieve current span from propagated request context
+        parent_span = get_current_span()
+
+        # start a new child span or a brand new trace if no parent
+        if parent_span is None:
+            span = tracer.start_trace(operation_name=operation)
+        else:
+            span = parent_span.start_child(operation_name=operation)
+        span.set_tag('server.http.url', request.full_url)
```
adriancole: ugh.. s/server.http.url/http.url/ please I understand why you are saying this, but ex in zipkin if there's actually an clash, it is tagged with an endpoint which compensates. On single-endpoint spans, this won't clash. This is likely to be read by people and I'd hate to make them stop and think about this part too much, or worse use it.

use-cases.md
```diff
@@ -0,0 +1,184 @@
...
+            span.log_event_with_payload('exception', e)
+            span.finish()
+            raise
+```
+
+  * The `get_current_span()` function is not a part of the OpenTracing API. It is meant to represent some util method of retrieving the current span from the current request context propagated implicitly (as is often the case in Python).
+  * The encoding function `trace_context_to_text` returns two maps, one representing the encoding of the trace context identity, the other the trace context attributes. We copy both maps into the HTTP request headers.
+  * We assume the HTTP client is asynchronous, so it returns a Future, and we need to add an on-completion callback to be able to finish the current child span.
+  * If HTTP client returns a future with exception, we log the exception to the span with `log_event_with_payload` method.
+  * Because the HTTP client may throw an exception even before returning a Future, we use a try/catch block to ensure that we finish the span in all circumstances.
```
adriancole: maybe add... to ensure it is reported and avoid leaking resources. Ex. this is commonly the reason, but that might not be obvious.

use-cases.md
adriancole: I think the example of auth-token is an unnecessarily scary one. Also, this assumes the primary context propagation api is OpenTracing, when in reality many frameworks have one to do things like this directly. I agree we should use an example more likely for a user to call, or at least not inviting discussions about security (by choosing a different field).

use-cases.md
```diff
@@ -0,0 +1,184 @@
...
+```
+
+### Logging Events
+
+We have already used `log_event_with_payload` in the client span use case. Events can be logged without a payload, and not just where the span is being created / finished. For example, the application may record a cache miss event in the middle of execution, as long as it can get access to the current span from request context:
+
+```python
+
+    span = get_current_span()
+    span.log_event('cache-miss')
```
adriancole: sweet. cc @marcingrzejszczak

use-cases.md
```diff
@@ -0,0 +1,184 @@
...
+
+### Setting "Debug" Mode
+
+Most tracing systems apply sampling to minimize the amount of trace data sent to the system.  Sometimes developers want to have a way to ensure that a particular trace is going to be recorded (sampled) by the tracing system, e.g. by including a special parameter in the HTTP request, like `debug=true`. The OpenTracing API does not have any insight into sampling techniques used by the implementation, so there is no explicit API to force it. However, the implementations are advised to recognized the `debug` Trace Attribute and take measures to record the span. In order to pass this attribute to tracing systems that rely on pre-trace sampling, the following approach can be used:
+
+```python
+
+    if request.get('debug'):
+        trace_context = tracer.new_root_trace_context()
+        trace_context.set_attribute('debug', True)
```
adriancole: I suggest changing this to sampled flag, not debug. The latter is very rarely implemented, and quite complex compared to the former.

cc @kristofa @marcingrzejszczak 

This topic is simultaneously quite important, and rarely implemented the same way. The debug flag is complex and only differs from the sampled flag in so far as it is passed to the collection tier. In many cases, collection tier is accepting 100%, so this complexity isn't warranted. As such, Brave doesn't even implementing this. If it were, you could imagine that collectors could choose to not honor debug anyway.

In brave, `X-B3-Sampled` is read regardless of whether an id is present in the request, and ensures that it is sampled by instrumentation

In Finagle, `X-B3-Sampled` is only read if there is an ID header. `X-B3-Flags` is read even if there's no id (and only includes debug). If the debug flag is set, it tripe `X-B3-Sampled` to 1, and is propagated in-band as debug *and also out-of-band* as span.debug = true.

use-cases.md
bcronin: Also +1 to combining start_trace and start_child

use-cases.md
yurishkuro: In my mind the main difference is that explicit propagation is done by the end user, and the business application needs to have all its method accept a context object. With implicit propagation the application does not need to change to have the propagation, because it is done by "magic" in a platform-specific way. Implicit propagation always has a performance penalty, while explicit propagation leaks an infrastructure concern into the application code. And in some languages like Go, implicit propagation is simply not possible (horrible hacks of identifying goroutine by the concatenation of its stack trace notwithstanding).

Open to suggestions on how to clarify all of this.

use-cases.md
yurishkuro: it is the policy of this office not to comment on ongoing investigations :-)

I wanted these examples to reflect the current API and to be used as a benchmark for proposals to change the API.

use-cases.md
adriancole: Based on what you said, I think the only thing that could clarify better
than what you have already is to switch the order.. implicit first, then in
the explicit section note that in some languages like go, implicit is not
possible.

That focuses a little better as I think I was reacting more to the nuances
of implicit things vs the idea that in some cases you can't do implicit
(even if you want to).

use-cases.md
yurishkuro: My intention here was to provide a way for application code to mark the trace as "debug" (whatever that means) in a vendor-agnostic way. Of course, the primary meaning I wanted was to ensure the trace is sampled, but OpenTracing API itself never talks about sampling anywhere, leaving it as an implementation detail of the underlying tracing system. So I am reluctant to use the sampling reference as the standard tag name, whereas `debug` can be such a standard tag that all tracers can recognize and handle in a special way.

BTW, in our internal Zipkin-compatible implementation the debug flag is used to raise both DEBUG and SAMPLED bits in the Flags, the former to persist in storage and potentially display differently in the UI, the latter to override sampling decision.

@bensigelman do you have thoughts on ^^^?

adriancole: thanks for clarifying your intent. makes sense to me

use-cases.md
yurishkuro: auth-token is something that was explicitly raised as a use case for distributed context propagation. On that, and on the scope of propagation functions in OT, please see https://github.com/opentracing/opentracing-go/issues/33

adriancole: Commented there.

use-cases.md
michaelsembwever: DEBUG isn't intuitive for me either. I understand the SAMPLED isn't either.

What about FORCE or FORCE_TRACING ?
Because *debug* does imply something else.

adriancole: >
> DEBUG isn't intuitive for me either. I understand the SAMPLED isn't either.
>
> What about FORCE or FORCE_TRACING ?
> Because *debug* does imply something else.
>
Yuri did mention a use case I hadn't thought about, which was to highlight
visually "Debug" traces (since the state is persisted). Debug may not be
the right word, but maybe force doesn't imply what he meant?

At any rate, it seems we are choosing words that aren't defined in OT as an
alternative to words also not defined in OT :P

yurishkuro: Ironically, the proposed API changes in [PR40](https://github.com/opentracing/opentracing-go/pull/40) make the currently documented approach invalid anyway, so we can mark it as ["open unsolved"](http://harrybosch.wikia.com/wiki/Open-Unsolved_Unit). Once PR40 is approved, we'll need to find a new approach, which is likely not to involve using tags to force the sampling, but rather a method parameter.

"Force tracing" sgtm.

dankosaur: Yeah let's defer this part of the decision.  Another option that is more flexible for end-tracers would be something like a "Trace Priority" flag or level.

PR #35, opened 2016-01-23
Add clarifying use case for why the TraceContext codec is part of the standard

bcronin:
Small proposed addition to explicitly clarify why a TraceContext codec is required in the standard.  

It calls out the obvious-in-retrospect case of needing to keep middleware generic that I hadn't been considering when initially trying to reason why the codec wasn't an implementation detail!

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -120,8 +120,9 @@ Implementations create `TraceContext` instances via a `TraceContextSource` inter
 - Create a new parent-less ("root") `TraceContext`. **(py: `new_root_trace_context`, go: `NewRootTraceContext`)**
 - Create a new child `TraceContext` given a parent `TraceContext`. **(py: `new_child_trace_context`, go: `NewChildTraceContext`)**
 
+Additionally, a `TraceContextSource` must be able to encode and decode `TraceContext`s. These interfaces are exposed so that reusable libraries and middleware can transport trace information opaquely without binding to the implementation details of any specific tracing system.
 
-Additionally, a `TraceContextSource` must be able to encode and decode `TraceContext`s. In its encoded form, a `TraceContext` is separated into a *pair* of values: one represents the "span identity" – for example, in Zipkin or Dapper, this would be the `trace_id` and `span_id`; and the other encoded value represents the trace attributes. This separation of encoded state enables optimizations in certain binary protocols. For instance, if a particular `TraceContext` has a fixed number of fixed-length "span identity" fields, then the span identity can be encoded into a fixed-length buffer (while of course the trace attributes cannot).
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -120,8 +120,9 @@ Implementations create `TraceContext` instances via a `TraceContextSource` inter
...
 
-Additionally, a `TraceContextSource` must be able to encode and decode `TraceContext`s. In its encoded form, a `TraceContext` is separated into a *pair* of values: one represents the "span identity" – for example, in Zipkin or Dapper, this would be the `trace_id` and `span_id`; and the other encoded value represents the trace attributes. This separation of encoded state enables optimizations in certain binary protocols. For instance, if a particular `TraceContext` has a fixed number of fixed-length "span identity" fields, then the span identity can be encoded into a fixed-length buffer (while of course the trace attributes cannot).
+In its encoded form, a `TraceContext` is separated into a *pair* of values: one represents the "span identity" – for example, in Zipkin or Dapper, this would be the `trace_id` and `span_id`; and the other encoded value represents the trace attributes. This separation of encoded state enables optimizations in certain binary protocols. For instance, if a particular `TraceContext` has a fixed number of fixed-length "span identity" fields, then the span identity can be encoded into a fixed-length buffer (while of course the trace attributes cannot).
```
bcronin: (To clarify, none of the text in this paragraph is changed: I just added a sentence after what had been the first sentence and added a paragraph break.)

yurishkuro:
:ship:

michaelsembwever:
@bcronin I still can't see the need for the codec in the application code.

Can you provide an example where application code needs to use this codec API?

bcronin:
An artificial, specific example would be this:

Alice wants to write a new open source, Thrift-like remote procedural call library. Among the protocols her library is built on are web sockets (streaming binary) and SSE (text-only). She wants to provide to use the OpenTracing API so consumers of her library get information-rich traces by using her library. She does not want to bind her library to any particular OpenTracing implementation; i.e. tracing data about what happens in her library should be there whether Zipkin, AppDash, or some tracer she’s never heard of it being used. However, since it is an RPC library, she needs some way to encode/serialize/marshal the context of the current trace on one side of the RPC and then decode/deserialize/unmarshal it on the other side of the RPC. She really does not know (or want to) concern herself with whether that “context” is just a simple TraceID int or a more complex set of parameters that the implementation wants to get across the wire to allow its approach to tracing to work. She uses the text codec for when she’s using SSE and uses the binary for everything else with assumption the binary form is more efficient.

michaelsembwever:
@bcronin, my misunderstanding to this has been based on my interpretation of what serialisation entails, and therefore my concerns amount to nothing more than poor naming to this "codec" part of the proposed API.

Serialisation is the writing of bytes or characters into streams/buffers for the sake of storage or cross-process communication.

While serialisation is obviously a consequence to this behind the scenes, it's not actually what the user cares about here, and therefore it's not intuitive API.

What this is really about is the transforming the state from the protocol's object to the tracing's context. There's no direct writing to bytes or characters, no writing to buffers or streams, no dealing with codecs.

While some protocol middleware frameworks will expose their underlying streams/buffers, and you'll be able to use the API's encoder/decoder to it, many won't and you'll be left having to transform objects from what the middleware provides into the OpenTracing's Span/TraceContext.

I've written these responses already in https://github.com/opentracing/opentracing.github.io/pull/34/files#r50616999 and https://github.com/opentracing/opentracing-java/pull/7/files#r50617464

PR #37, opened 2016-01-24
Clarify two context propagation techniques; address some comments.

yurishkuro:
@adriancole addresses some of your comments.


adriancole:
LGTM thanks for the follow-up

PR #38, opened 2016-01-24
Replace "serialization' with "encoding"

yurishkuro:


michaelsembwever:
was there any discussion to this?
I presume there was, but could you link to the discussion please.
It's nice to be able to see and re-read the rationale for changes made.

yurishkuro:
@michaelsembwever was done based on objections in several threads to the use of the word "serialization". None of the classes are actually called that, so this change lines up documentation with the class names.

Issue #24, opened 2016-01-15
consolidate the myriad interfaces into Tracer


bensigelman:
@bcronin and I were chatting this AM (we sit next to each other :), and there's a case for collapsing more of these interfaces into Tracer and just using loud comments/etc to delineate the boundaries between the pieces that have to do with span-starting vs the pieces that create TraceContexts vs the pieces that do codec of same. Inheritance in interfaces – especially when the bottom line is that Tracer implements essentially everything – can make things harder to grok, even if our intention is to honor a separation of concerns.

On #21, @yurishkuro added:

> There are definitely benefits in keeping everything that Tracer can do in one place. Especially for dynamic languages - I was able to kludge Python version with multiple inheritance, but it was weird, and in Javascript - phhhhh... So +1 to that too.

slimsag:
FWIW I agree with this, but I'm not sure of the details exactly. I'd like to see it in code before giving my final opinion.

adriancole:
Heh looks like this topic came up in 3 places! Short of it is that we want
things to be easy to grok, and clean in the target language. In some cases,
like java, composition will be far more preferred over inheritance.

Each part is simpler to understand and multiple inheritance for orthogonal
concerns is a deprecated practice (composition is the current way)

Iotw something doesnt have to be a codec to start traces! The actual call
sites are orthogonal, and mixing all concerns in an uber interface can
cause some languages (like java) to be nastier to use and implement.

slimsag:
I think @adriancole raises an interesting point, maybe we shouldn't strive for a 1:1 API across all languages but instead for 1:1 _features_ and write it in the language however you normally would, e.g. in Go using composition instead of a monolithic interface, and in Python and Javascript using a single Tracer object / in one place if that's most idiomatic to that language.

Thoughts @yurishkuro @bcronin ?

bensigelman:
@slimsag I'm 100% (110%? What's my allowance?) in favor of semantic parity rather than API parity... that's the whole point of the intentionally-prose-rather-than-pseudocode "semantic spec" page we've been working through.

bcronin:
For what it's worth, my metric is simply if a user of OpenTracing is familiar with it in language A, it should be intuitive for her to use OpenTracing language B. That's pretty flexible constraint, so I think there are a lot of different approaches that could work.

[...]

PR #42, opened 2016-01-27
remove TraceContext from the docs, too

bensigelman:
Per https://github.com/opentracing/opentracing-go/pull/40 and https://github.com/opentracing/opentracing.github.io/issues/24

I'll add a few comments inline.

This and https://github.com/opentracing/opentracing-go/pull/40 should have a shared fate.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -124,7 +124,7 @@ Note that there is no expectation that different tracing systems propagate `Span
 The `SpanPropagator` interface must have the following capabilities:
 - Propagating a Span instance by encoding it as either...
   - a pair of binary values **(py: `propagate_span_as_binary`, go: `PropagateSpanAsBinary`)**, or
-  - a pair of text-encoded `string->string` maps, where the keys are suitable for use in HTTP headers (see the notes about "trace attribute" keys above) **(py: `propagate_span_as_text`, go: `PropagateSpanAsText`)**
+  - a pair of unicode `string->string` maps, where the keys are suitable for use in HTTP headers (see the notes about "trace attribute" keys above) **(py: `propagate_span_as_text`, go: `PropagateSpanAsText`)**
 - Returning a new Span instance that joins to a Span previously propagated via either...
   - a pair of binary values **(py: `join_trace_from_binary`, go: `JoinTraceFromBinary`)**, or
-  - a pair of `string->string` maps **(py: `join_trace_from_text`, go: `JoinTraceFromText`)**
+  - a pair of unicode `string->string` maps **(py: `join_trace_from_text`, go: `JoinTraceFromText`)**
--- a/spec.md
+++ b/spec.md
@@ -49,9 +49,9 @@ Every Span has zero or more **Logs**, each of which being a timestamped event na
 
 Every Span may also have zero or more key/value **Tags**, which do not have timestamps and simply annotate the spans.
 
-Every Span is bound to a **Trace Context**. The Trace Context describes how the Span it's bound to fits in to the larger Trace. A Trace Context encapsulates the smallest amount of state needed to continue a Trace across a process boundary (or, more generally, a serialization boundary). Put another way, a Span cannot exist without a Trace Context, but a Trace Context can be (and is) serialized without a Span: Spans are built around their Trace Context and concern themselves with application-level instrumentation (logs, timing information, tags). Note that, in order to support *Trace Attributes* (see below) that propagate across a distributed Trace transparently, those attributes must be represented by/within the Trace Context.
+Spans may be **propagated**; that is, they may cross process boundaries along with sufficient information to continue the Trace on the far side of the propagation.
 
-**Trace Attributes** are key/value pairs stored in a Trace Context and propagated _in-band_ to all future child Spans. Given a full-stack OpenTracing integration, Trace Attributes enable powerful functionality by transparently propagating arbitrary application data: for example, an end-user id may be added as a Trace Attribute in a mobile app, propagate (via the distributed tracing machinery) into the depths of a storage system, and recovered at the bottom of the stack to identify a particularly expensive SQL query.
+**Trace Attributes** are key/value pairs stored in a Span and propagated _in-band_ to all future child Spans. Given a full-stack OpenTracing integration, Trace Attributes enable powerful functionality by transparently propagating arbitrary application data: for example, an end-user id may be added as a Trace Attribute in a mobile app, propagate (via the distributed tracing machinery) into the depths of a storage system, and recovered at the bottom of the stack to identify a particularly expensive SQL query.
 
 Trace Attributes come with powerful _costs_ as well; since the attributes are propagated in-band, if they are too large or too numerous they may reduce system throughput or contribute to RPC latency.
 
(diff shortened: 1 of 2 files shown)


spec.md
```diff
@@ -49,9 +49,9 @@ Every Span has zero or more **Logs**, each of which being a timestamped event na
 
 Every Span may also have zero or more key/value **Tags**, which do not have timestamps and simply annotate the spans.
 
-Every Span is bound to a **Trace Context**. The Trace Context describes how the Span it's bound to fits in to the larger Trace. A Trace Context encapsulates the smallest amount of state needed to continue a Trace across a process boundary (or, more generally, a serialization boundary). Put another way, a Span cannot exist without a Trace Context, but a Trace Context can be (and is) serialized without a Span: Spans are built around their Trace Context and concern themselves with application-level instrumentation (logs, timing information, tags). Note that, in order to support *Trace Attributes* (see below) that propagate across a distributed Trace transparently, those attributes must be represented by/within the Trace Context.
+Spans may be **propagated**; that is, they may cross process boundaries along with sufficient information to continue the Trace on the far side of the propagation.
```
bensigelman: The number of concepts and the number of things we need to explain on the right side of this diff are smaller and simpler than the left side: probably a good sign.

spec.md
```diff
@@ -111,26 +94,37 @@ In any case, the programmatic `TraceContext` interface is deceptively simple:
...
 
-A complete `Tracer` implementation must also satisfy the requirements of the `TraceContextSource`, `TraceContextEncoder`, and `TraceContextDecoder` interfaces.
+### Tracer
+
+The `Tracer` interface must have the following capabilities:
+
+- Start a `Span` that has no parent, commonly referred to as a *root* `Span` **(py: `start_trace`, go: `StartTrace`)**
+- Provide some form of access to a SpanPropagator (whether that's through embedding, inheritance, containment, or something else is per-platform)
+
+### SpanPropagator
```
bensigelman: most of the below copy is lifted and adapted from the golang comments I wrote.

use-cases.md
```diff
@@ -192,22 +193,24 @@ We have already used `log_event_with_payload` in the client span use case. Event
...
-```python
-
-    if request.get('debug'):
-        trace_context = tracer.new_root_trace_context()
-        trace_context.set_attribute('debug', True)
-        span = tracer.start_span_with_context(
-            operation_name=operation, 
-            trace_context=trace_context)
-```
+`XXX: we don't have an API for this anymore.`
```
bensigelman: (a known issue that *could* be addressed if we deem it necessary... @yurishkuro)

use-cases.md
```diff
@@ -44,41 +44,43 @@ These two examples are intentionally naive. Usually developers will not want to
...
+The `span` object above can be `None` if the Tracer did not find relevant headers in the incoming request: presumably because the client did not send them. In this case the server needs to start a brand new trace.
 
 ```python
-    if context is None:
+    span = tracer.join_trace_from_text(
+        operation_name=operation,
+        context_snapshot=request.headers,
+        trace_attributes=request.headers
+    )
+    if span is None:
```
bensigelman: this little snippet seems needlessly awkward to me... when would you try to join with a trace and *not* just create the root span of the same name in the failure scenario?

spec.md
```diff
@@ -111,26 +94,37 @@ In any case, the programmatic `TraceContext` interface is deceptively simple:
...
+A SpanPropagator is responsible (a) for encoding Span instances in a manner suitable for propagation, and (b) for taking that encoded data and using it
+to generate Span instances that are placed appropriately in the overarching Trace. Typically the propagation will take place across an RPC boundary, but message queues and other IPC mechanisms are also reasonable places to use a SpanPropagator.
 
-- Create a new parent-less ("root") `TraceContext`. **(py: `new_root_trace_context`, go: `NewRootTraceContext`)**
-- Create a new child `TraceContext` given a parent `TraceContext`. **(py: `new_child_trace_context`, go: `NewChildTraceContext`)**
+The encoded form of a propagated span is divided into two components:
 
-Additionally, a `TraceContextSource` must be able to encode and decode `TraceContext`s. These interfaces are exposed so that reusable libraries and middleware can transport trace information opaquely without binding to the implementation details of any specific tracing system.
+1. The core identifying information for the Span (for example, in Dapper this would include a `trace_id`, a `span_id`, and a bitmask representing the sampling status for the given trace)
```
yurishkuro: we may need to introduce "context snapshot" as a concept

use-cases.md
yurishkuro: don't know of real life examples. An imaginary example could be some middleware / routing layer that only wants to participate in traces, but never initiate them.

yurishkuro: in fact, I remember it came up once with Hyperbahn/TChannel, where Hyperbahn would peek into the trace ID, and if present it would start a new span and replace the trace id in the message being routed (that's the reason why trace ID was stored in a known fixed offset in the message, not in the app headers), but if there is no trace ID or it's not sampled, it would leave it alone.

use-cases.md
```diff
@@ -192,22 +193,24 @@ We have already used `log_event_with_payload` in the client span use case. Event
     span.log_event('cache-miss') 
 ```
 
-The tracer automatically records a timestamp of the event, in contrast with tags that apply to the entire span. In the future versions of the API it will be possible to associate an externally provided timestamp with the event, e.g. see [pull request #38](https://github.com/opentracing/opentracing-go/pull/38).
+The tracer automatically records a timestamp of the event, in contrast with tags that apply to the entire Span. In the future versions of the API it will be possible to associate an externally provided timestamp with the event, e.g. see [pull request #38](https://github.com/opentracing/opentracing-go/pull/38).
```
yurishkuro: we can remove PR reference, as it's been merged, and just do an example of generic log method

yurishkuro:
LGTM

use-cases.md
bensigelman: Done.

spec.md
bensigelman: it's a B- name at best... I added something, though.

adriancole:
Definitely positive as some of the ancillary features are put in more context (no pun intended), and agreed.. Less things to explain overall.

I dare ask... Context Snapshot? Sounds like we are still struggling on naming the fixed-length, mandatory data from the nice to have data.

bensigelman:
@adriancole: struggle++

Suggestions very welcome.

On Wednesday, January 27, 2016, Adrian Cole <notifications@github.com>
wrote:

> Definitely positive as some of the ancillary features are put in more
> context (no pun intended), and agreed.. Less things to explain overall.
>
> I dare ask... Context Snapshot? Sounds like we are still struggling on
> naming the fixed-length, mandatory data from the nice to have data.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/opentracing/opentracing.github.io/pull/42#issuecomment-175697887>
> .
>


-- 
(Sent via mobile)

adriancole:
Thought about this last night.

One of the struggles is about how to contrast the two propagated data. What
if it was classified by who's responsible vs its use? Ex. Tracer Attributes
vs User Attributes

The last 'r' in Tracer Attributes isn't a typo, basically it is the
responsibility of the tracer to define and control the former data. It is
SPI

A user is who is responsible for the latter data, both knowing what it is
for and defining how to use it. It is API

There's some precedence for this.

In many cloud apis, tags are called "user metadata" and I've seen in at
least one storage system (atmos I think) "system metadata", which coexists
in with user metadata, but it is the system's responsibility to manage.

yurishkuro:
I like that.

michaelsembwever:
:+1: 

**Removing TraceContext definitely made the specification a lot easier to read.**

(I still think there's room for more! 
 - moving functionality on spans out of Span,
 - SpanPropagator still comes across as unnecessarily heavy,
 - the idea of binary and text flavours seems to take an unneeded jump down to the wire protocol

bensigelman:
@michaelsembwever yes, agreed that there is still more to discuss with the group. Per gitter, I wanted to get in the TraceContextEctomy before proceeding to these other topics. I am having personal queuing problems this week but – if all goes well – I am hoping to do triage and organization/prioritization of open issues.

Issue #44, opened 2016-02-01
meta-issue: stabilize OpenTracing 1.0RC semantics


bensigelman:
I mentioned this in my [OpenTracing status report](https://docs.google.com/presentation/d/1Eexa7lZRCn7KE0w_4I0lZW99l2d-p7YbVyKylxWGz0Y/edit#slide=id.g108c88bc6a_0_18), but to reiterate here...

This is the rough sequence of steps before I would feel ok declaring an OpenTracing 1.0 semantic spec:

1. Agree that there’s a problem to solve
1. Agree on the basic approach / vision
1. Propose initial semantic spec and specific platform sketches, with examples
1. Converge on a “1.0RC” semantic specification (with rough implementations on multiple platforms)
1. Allow early-adopters to bind to the “1.0RC” spec (app programmers, IPC frameworks, control flow systems, and of course tracing implementations)
1. Converge on a “1.0” semantic spec and announce more broadly, "co-branded" with various OpenTracing users and implementations

Right now we are somewhere between (3) and (4) above, and this particular Issue is about getting to (4).

Here are the more controversial issues I'm remembering, some of which I think are mostly resolved:
- [x] Should DCP be modeled separately from Spans/etc? (Current answer: “no”)
- [x] Should we support a robust message logging API? (Current answer: “maybe, but not yet”)
- [x] Should we support (auto-propagated) “Trace Attributes”? (Current answer: “yes, with copious warnings in comments”)
- [x] How should the Propagator interface interoperate with IPC frameworks? (Undecided)
- [x] The best API for child-Span creation (Undecided)

There are also some lingering naming concerns, but I'm confident we can get through those if we deal with the semantics.

I would further suggest that we try to hash out the remaining sticking points via a google hangout rather than on Github. On that note, please fill out:

   http://doodle.com/poll/q82qiz932q9gm3dr

**If** we need to use Github due to timezone sadness, I would ask that we restrict ourselves to issues in [opentracing.github.io](https://github.com/opentracing/opentracing.github.io/issues) (creating new ones as needed) rather than fragmenting the discussions across per-language repos. And concerns about the current semantic spec should be accompanied by *specific* suggestions to the relevant [use cases](http://opentracing.io/use-cases/).

In alphabetical order: I suspect that some of @adriancole @bcronin @dkuebric @michaelsembwever @slimsag @yurishkuro might care about this... of course all are welcome to chime in!

(^^^ alphabetical order on the github usernames, btw :)

[...]

PR #45, opened 2016-02-02
Simplify the SpanPropagator to SpanInjector and SpanExtractor.

michaelsembwever:
Simplify the SpanPropagator to SpanInjector and SpanExtractor.
 - works with the bare minimum we know (given a middleware object we need to inject
and extract a span to/from it)
 - permits implementations of these to deal with complexities…

ref:
 - https://github.com/opentracing/opentracing-java/pull/8#discussion_r51540744

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -100,31 +100,14 @@ The `Span` interface must have the following capabilities:
 The `Tracer` interface must have the following capabilities:
 
 - Start a `Span` that has no parent, commonly referred to as a *root* `Span` **(py: `start_trace`, go: `StartTrace`)**
-- Provide some form of access to a SpanPropagator (whether that's through embedding, inheritance, containment, or something else is per-platform)
+- Provide access to a SpanInjector and SpanExtractor that's appropriate for a specified IPC system, or a helper implementation that can be delegated to. **(py: `get_span_injector, get_span_extractor`, go: `GetSpanInjector, GetSpanExtractor`)**
 
-### SpanPropagator
+### SpanInjector
 
-A SpanPropagator is responsible (a) for encoding Span instances in a manner suitable for propagation, and (b) for taking that encoded data and using it
-to generate Span instances that are placed appropriately in the overarching Trace. Typically the propagation will take place across an RPC boundary, but message queues and other IPC mechanisms are also reasonable places to use a SpanPropagator.
+The `SpanInjector` interface must have the following capabilities:
+ - inject the span into the middleware object **(py: `inject`, go: `inject`)**
 
-The encoded form of a propagated span is divided into two components:
+### SpanExtractor
 
-1. The core identifying information for the Span, referred to as the "context snapshot" (for example, in Dapper this would include a `trace_id`, a `span_id`, and a bitmask representing the sampling status for the given trace)
-1. Any trace attributes (per Span's ability to set trace attributes that propagate across process boundaries)
-
-The encoded data is separated in this way for a variety of reasons; the most important is to give OpenTracing users a portable way to opt out of Trace Attribute propagation entirely if they deem it a stability risk.
-
-The propagation and trace-joining methods come in two flavors: binary and text.
-
-- The *text* format is a platform-idiomatic map from (unicode) `string` to `string`; it is better-suited to pretty-printing and debugging
-- The *binary* format is an opaque byte array and is better-suited to compact, high-performance encoding, decoding, and transmission
-
-Note that there is no expectation that different tracing systems propagate `Spans` in compatible ways. Though OpenTracing is agnostic about the tracing implementation, for successful inter-process handoff it's essential that the processes on both sides of a propagation use the same tracing implementation.
-
-The `SpanPropagator` interface must have the following capabilities:
-- Propagating a Span instance by encoding it as either...
-  - a pair of binary values **(py: `propagate_span_as_binary`, go: `PropagateSpanAsBinary`)**, or
-  - a pair of unicode `string->string` maps, where the keys are suitable for use in HTTP headers (see the notes about "trace attribute" keys above) **(py: `propagate_span_as_text`, go: `PropagateSpanAsText`)**
-- Returning a new Span instance that joins to a Span previously propagated via either...
-  - a pair of binary values **(py: `join_trace_from_binary`, go: `JoinTraceFromBinary`)**, or
-  - a pair of unicode `string->string` maps **(py: `join_trace_from_text`, go: `JoinTraceFromText`)**
+The `SpanExtractor` interface must have the following capabilities:
+ - extract the span from the middleware object **(py: `extract`, go: `extract`)**
--- a/use-cases.md
+++ b/use-cases.md
@@ -53,11 +53,7 @@ When a server wants to trace execution of a request, it generally needs to go th
 Let's assume that we have an HTTP server, and the Span is propagated from the client via HTTP headers, accessible via `request.headers`:
 
 ```python
-    span = tracer.join_trace_from_text(
-        operation_name=operation,
-        context_snapshot=request.headers,
-        trace_attributes=request.headers
-    )
+    span = tracer.get_span_extractor(request.class).extract(request)
 ```
 
 Here we set both arguments of the decoding method to the `headers` map. The Tracer object knows which headers it needs to read in order to reconstruct the context snapshot and any Trace Attributes.
@@ -69,11 +65,7 @@ The `operation` above refers to the name the server wants to give to the Span. F
 The `span` object above can be `None` if the Tracer did not find relevant headers in the incoming request: presumably because the client did not send them. In this case the server needs to start a brand new trace.
 
 ```python
-    span = tracer.join_trace_from_text(
-        operation_name=operation,
-        context_snapshot=request.headers,
-        trace_attributes=request.headers
-    )
+    span = tracer.get_span_extractor(request.class).extract(request)
     if span is None:
         span = tracer.start_trace(operation_name=operation)
     span.set_tag('http.method', request.method)
@@ -122,7 +114,7 @@ The downside of explicit context propagation is that it leaks what could be cons
 
 ### Tracing Client Calls
 
-When an application acts as an RPC client, it is expected to start a new tracing Span before making an outgoing request, and propagate the new Span along with that request. The following example shows how it can be done for an HTTP request. 
+When an application acts as an RPC client, it is expected to start a new tracing Span before making an outgoing request, and propagate the new Span along with that request. The following example shows how it can be done for an HTTP request.
 
 ```python
     def traced_request(request, operation, http_client):
@@ -137,15 +129,9 @@ When an application acts as an RPC client, it is expected to start a new tracing
         span.set_tag('http.url', request.full_url)
 
         # Propagate the Span via HTTP request headers
-        h_ctx, h_attr = tracer.propagate_span_as_text(span)
+        tracer.get_span_injector(request.class).inject(span, request)
 
-        for key, value in h_ctx.iteritems():
-            request.add_header(key, value)
-        if h_attr:
-            for key, value in h_attr.iteritems():
-                request.add_header(key, value)
-
-        # define a callback where we can finish the span 
+        # define a callback where we can finish the span
         def on_done(future):
             if future.exception():
                 span.log_event_with_payload('exception', exception)
@@ -190,7 +176,7 @@ We have already used `log_event_with_payload` in the client Span use case. Event
 ```python
 
     span = get_current_span()
-    span.log_event('cache-miss') 
+    span.log_event('cache-miss')
 ```
 
 The tracer automatically records a timestamp of the event, in contrast with tags that apply to the entire Span. It is also possible to associate an externally provided timestamp with the event, e.g. see [Log (Go)](https://github.com/opentracing/opentracing-go/blob/ca5c92cf/span.go#L53).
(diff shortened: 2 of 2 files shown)


use-cases.md
```diff
@@ -53,11 +53,7 @@ When a server wants to trace execution of a request, it generally needs to go th
 Let's assume that we have an HTTP server, and the Span is propagated from the client via HTTP headers, accessible via `request.headers`:
 
 ```python
-    span = tracer.join_trace_from_text(
-        operation_name=operation,
-        context_snapshot=request.headers,
-        trace_attributes=request.headers
-    )
+    span = tracer.get_span_extractor(request.class).extract(request)
```
yurishkuro: imo this makes the OT implementation work much harder than it needs to. For the end user it looks like magic, and implementation needs to know about 100s of various rpc request types. I don't think it's a useful way to spend implementors time chasing after all different ways RPC frameworks express their requests and learning those APIs, not to mention that the classpath dependency of a tracer will become huge.

In the current API, the OT implementor only needs to support two representations, a binary blob and a string key/value map. End users (RPC implementors) can use those two representations and inject them into their own Request class using their own API, without making it the responsibility of the OT tracer.

yurishkuro: Another issue is that it makes the tracer very inflexible. If tomorrow you decide to create your own MicksSuperFastRpcOverHttp2 framework, you won’t be able to use any of the existing OT tracers to instrument that framework, because none of them will know about MicksSuperFastHttpRequest API. Where as in the current API, any framework can use OT to instrument itself.

What I do like about this direction is an implication that the data can be written into the request potentially more efficiently than by allocating and returning two maps, then iterating through then and writing them into yet another map (headers). @bensigelman I think we should consider building better streaming support into the codec API. For example, in Go version, instead of allocating and returning two byte arrays, we could’ve taken a Writer as an argument and just write the bytes to it. That’s much more in line with std Go APIs. If people, for some reason, must get a map or blobs back, we can always provide simple wrappers to do so. For example we could define the codec as taking this as argument

```go
type BinarySpanWriter interface {
    WriteSpanID(bytes []byte)
    WriteTraceAttributes(bytes byte[])
}

type KVMapSpanWriter interface {
    WriteSpanIDKV(key, value string)
    WriteTraceAttribute(key, value string)
}
```

On the reading side, it would be done with an interator-like interface. I can create concrete examples in a separate PR.

michaelsembwever:
We do not know if all RPC systems will work with a fixed choice between byte[] or Map<String,String>.

Furthermore, per https://github.com/opentracing/opentracing-java/pull/8#discussion_r51540744

The only reason the separation of trace identity and trace attributes exists is
**if** the middleware exposes a binary stream and is using binary serialisation
**if** the binary protocol uses fixed-length slots,
**if** the IPC interaction wants to skip trace attributes
**for** a small performance benefit or stability fix.

The design of `SpanPropagator` was targeted toward niche performance optimisations for the additional DCP add-on (or the lack of the Specification to lay clear guidelines to the intended use of DCP). I fear our API design is going to quickly become fragile and brittle if we need to go to these extents for each specific thing, rather than designing toward commonalities and the general use-cases of tracing.

This API improvement instead intends to keep horizons as open as possible to cater for the adoption of as many tracing frameworks as possible.

use-cases.md
michaelsembwever: > imo this makes the OT implementation work much harder than it needs to. 

It doesn't. Your existing approaches are still there.

    tracer.get_span_injector(map.class).inject(span, map)
    span = tracer.get_span_extractor(map.class).extract(map)

    tracer.get_span_injector(byte_buffer.class).inject(span, byte_buffer)
    span = tracer.get_span_extractor(byte_buffer.class).extract(byte_buffer)

> For the end user it looks like magic, and implementation needs to know about 100s of various rpc request types. 

No it doesn't. Because standard helper implementations are provided.

For the popular rpc request types we can provide implementations, but these too should be rather thin re-using the helper implementations [eg  tracer.get_span_extractor(map.class)]

> the classpath dependency of a tracer will become huge.

No it won't. These dependencies will of course be marked as optional or provided.

> In the current API, the OT implementor only needs to support two representations, a binary blob and a string key/value map. 

This won't support all rpc request types, it won't be optimised for all rpc request types, and it ties the API down to these limitations.

yurishkuro: @michaelsembwever I misunderstood your proposal, I thought your `request.class` was something like `HttpRequest.class`.  If it's instead a basic data type line byte[] or map, then it's certainly doable (should be made explicit in the doc though). And the `request.class` can even be one of the streaming-oriented APIs I mentioned above, +1 for that. Another thing I like is that it abstracts away the separation of spanID + traceAttrs to a certain degree (although it's still a question how to do it in the binary form of this approach).

The main concern people will have with that is the same one as on the `setTag(String, Object)` signature. There's no guarantee that the end user won't supply some nonsense as an argument, and the tracer will not be able to propagate the context. The spec for the tracer would be requiring to support a certain minimal set of `request.class`es, and if it supports more, end users can't really rely on that in their code. So flexibility in this regard hurts portability.

spec.md
```diff
@@ -100,31 +100,14 @@ The `Span` interface must have the following capabilities:
 The `Tracer` interface must have the following capabilities:
 
 - Start a `Span` that has no parent, commonly referred to as a *root* `Span` **(py: `start_trace`, go: `StartTrace`)**
-- Provide some form of access to a SpanPropagator (whether that's through embedding, inheritance, containment, or something else is per-platform)
+- Provide access to a SpanInjector and SpanExtractor that's appropriate for a specified IPC system, or a helper implementation that can be delegated to. **(py: `get_span_injector, get_span_extractor`, go: `GetSpanInjector, GetSpanExtractor`)**
```
bensigelman: meta-point: like @yurishkuro, I, too, did not understand what you were getting at when I read this bullet... if we decide to proceed with this direction I would like to rewrite this to be a little clearer about the way that the injector/extractor impl is inferred.

use-cases.md
michaelsembwever: > I thought your request.class was something like HttpRequest.class

That is correct and is what I meant in the initial examples, but it's not restricted to specific rpc request types. The example of using map.class or byte_buffer.class was raised though to illustrate this.

An example of rpc request based injector/extractors being thin is…

    public class HttpRequestSpanInjector implements SpanInjector {
         public void inject(Span span, HttpServletRequest request) {
              tracer.getSpanInjector(Map.class).inject(span, request.getheaders());
        }
    }

> The main concern people will have with that is the same one as on the setTag(String, Object) signature. 

Yes. Throwing a NoInjectorFoundException would be normal part of the Specfication.

     // throws a NoInjectorFoundException
     tracer.getSpanInjector(SomethingCompletelyWrong.class)

> So flexibility in this regard hurts portability.

get_span_injector(map.class) and get_span_injector(byte_buffer.class) is intended to solve that problem (and application code can plug in its own injectors and extractors)

spec.md
michaelsembwever: I did keep it brief. Given ongoing discussions would influence things.

use-cases.md
```diff
@@ -69,11 +65,7 @@ The `operation` above refers to the name the server wants to give to the Span. F
 The `span` object above can be `None` if the Tracer did not find relevant headers in the incoming request: presumably because the client did not send them. In this case the server needs to start a brand new trace.
 
 ```python
-    span = tracer.join_trace_from_text(
-        operation_name=operation,
-        context_snapshot=request.headers,
-        trace_attributes=request.headers
-    )
+    span = tracer.get_span_extractor(request.class).extract(request)
```
bensigelman: (starting a different thread here since I don't want to muddy the discussion above with @yurishkuro)

@michaelsembwever I think this is an appealing direction and I hope we can do something along these lines. Of course nothing is ever so simple, though. :)  I do have a few concerns...

1. There is a "stuttering" aspect to these code snippets: namely that the type of an object is used to get an Injector/Extractor, then that object is subsequently passed in to the inject/extract method. One obvious alternative would be to consolidate around a single method for the common case... basically a `Tracer.extract(obj, extractor=None)` sort of signature. If `extractor` is unspecified, then `extract` makes its best guess based on `obj` and goes from there. I also prefer this approach because it doesn't force use into guessing the correct extractor based only on the type information: for instance, there might be more than one way to encode data into a `string->string` map.
1. Related to that last point, I like how this design allows the decision to keep/drop trace attributes a per-Injector/Extractor concern. So, again, with the sort of signature I've proposed above, we could have something like `Tracer.extract(request, extractor=StringMapExtractor(attrs=False))`.
1. Of course something would be wrong with the world if I didn't quibble about naming... and I would like to do so if we can come to agreement on the higher-order bits above.

bensigelman:
@michaelsembwever this proposal is a promising one and I appreciate that you put this together. I also wanted to state that, as a matter of process, thinking about this sort of proposal at the `spec.md` level was way more effective as far as my brain is concerned.

use-cases.md
bensigelman: Oh:

4) I initially wrote up OT interfaces that separated encoding/decoding (or injection/extraction, whatever). There were various requests for consolidation, mostly stemming from a desire to keep the number of OT types/interfaces to a minimum.

use-cases.md
yurishkuro: > application code can plug in its own injectors and extractors

only by delegating to some other "known" injectors provided by the tracer, as ultimately the app has no idea how to encode the distributed context into binary or kv-map.

michaelsembwever: > only by delegating to some other "known" injectors provided by the tracer, as ultimately the app has no idea how to encode the distributed context into binary or kv-map.

there's further journey to be had here, but this is going into more the internal API side of things where we have more freedom to make mistakes.

use-cases.md
michaelsembwever: 1) That's a possibility. I'm thinking though of using an Inversion of Control here so that Injector/Extractor implementations can claim responsibility of what objects that can handle, rather than the user having to know. A problem with the tracer extract method is it's a plumbing api (convenience method on top of another).

    tracer.extract(object)
    tracer.getSpanExtractor(object.class).extract(object)

2) This can still be done with my original proposal

    SpanExtractor custom = new StringMapExtractor(false);
    Span span = custom.extract(map);
Here the use of explicitly calling constructors intentionally takes us into the land of the internal APIs.

4) Yes there's no reason that Injector and Extractor don't sit in once interface. Just didn't have a good name for it, and wanted a little added clarity at this point.

bensigelman: There's one problem with (2) above – namely that the `StringMapExtractor` is defined in an OT implementation, and as such we run into portability problems. FWIW, my own suggestion had the same issue.

How do we consistently identify Extractors/Injectors – and, worse still, any parameters they take – without breaking the core OT interface abstraction? We could do something gross with an enum (one entry for each type of extractor+injector pair).

yurishkuro: you guys both lost me at first with `StringMapExtractor`, now Ben's comment highlights exactly my concern - end user cannot write any Extractors, the encoding details are private to the tracer.

michaelsembwever: >  end user cannot write any Extractors, the encoding details are private to the tracer.

Yes and No.

The end user can write Injector/Extractors, but yes they need to honour existing OT protocol-specific specification. This is tackled by providing base Injectors/Extractors for protocols and protocol-specific specifications. I don't think we should be bloating the primary OT Specification with protocol specifics, but rather provide separate Protocol Specifications. This too creates the clean separation between Public API and Internal APIs.

yurishkuro: I think what you're saying is that the custom end-user written codec will not be an `f(span, target)` but an `f(span, target, g)`, where `g = tracer.injector(map or binary)`, i.e. it will have to use other (more standard) injector from tracer into a known & shared data model, and then read from that model into the `target` which is unknown to the tracer.

```
func f(span, target) {
    g := tracer.injector(map[string]string)
    m := map()
    g.inject(span, m)
    // now do m => target
}
```

dkuebric:
I see the primary appeal of this as potential for performance optimization, skipping an intermediary representation in a lot of cases.  As such this this feels more like an optimization than a capability issue: it could be added later without any loss of backward compatibility.

use-cases.md
bensigelman: After sleeping on this, I would prefer that we use something more explicit than param-type-based inference to determine the appropriate injector/extractor for the given job. This could be via some sort of registry mechanism or it could be disgusting enum. But using the type alone seems fragile and "magical in a bad way."

I still like the core premise of this PR: namely, an "official" interface for injection/extraction (rather than the two ad hoc `SpanPropagator` methods we have right now) as it makes it easier to imagine adding support for particularly popular protocols at the OT layer... I'm thinking of large, polyglot, popular IPC mechanisms like grpc, thrift, tchannel, whatever.

@michaelsembwever it would be great to talk about this at our inaugural hangout this morning/afternoon/evening depending on your TZ :)

bensigelman:
@dkuebric ACK that this could be used as a performance optimization, and also ACK that it *could* be added later... but if we pursue this path we would want to deprecate the bare "binary" and "text" methods in SpanPropagator, so if we know we want to take this path we may as well do so now. Or that would be my argument.

use-cases.md
michaelsembwever: > I think what you're saying is that the custom end-user written codec will not be an f(span, target) but an f(span, target, g)  where g = tracer.injector(map or binary)

No. There is no need to add a parameter, especially since this is not a per call argument.
My presumption is that the injector already holds reference to the tracer, and therefore can obtain the injector to delegate to, by itself.

bensigelman:
cross-ref: https://github.com/opentracing/opentracing.github.io/issues/44#issuecomment-179654948

yurishkuro:
I think this has been addressed by recent PRs in Go (https://github.com/opentracing/opentracing-go/pull/48) and Python (https://github.com/opentracing/opentracing-python/pull/17).

If no objections, will close in 48hrs.

yurishkuro:
No objections received, so closing.

PR #46, opened 2016-02-02
Move functionality on Spans from existing within Span to Tracer.

michaelsembwever:
Move functionality on Spans from existing within Span to Tracer.
 - removes Span.create_child
 - removes Span.finish
 - adds Tracer.finish

ref:
 - https://github.com/opentracing/opentracing-java/commit/71f7510094f5c44a184cb7e54bb3d369c24fec3e

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -71,8 +71,6 @@ OpenTracing supports a number of different platforms, and of course the per-plat
 
 The `Span` interface must have the following capabilities:
 
-- Create and start a new child `Span` with a given operation name. Any trace attributes must be passed through into the child. **(py: `start_child`, go: `StartChild`)**
-- Finish the (already-started) `Span`.  Finish should be the last call made to any span instance, and to do otherwise leads to undefined behavior. **(py: `finish`, go: `Finish`)**
 - Set a key:value tag on the `Span`. The key must be a `string`, and the value must be either a `string`, a `boolean`, or a numeric type. Behavior for other value types is undefined. If multiple values are set to the same key (i.e., in multiple calls), implementation behavior is also undefined. **(py: `set_tag`, go: `SetTag`)**
 - Add a new log event to the `Span`, accepting an event name `string` and an optional structured payload argument. If specified, the payload argument may be of any type and arbitrary size, though implementations are not required to retain all payload arguments (or even all parts of all payload arguments). **(py: `log_event, log_event_with_payload`, go: `LogEvent, LogEventWithPayload`)**
 - Set a trace attribute, which is a simple string:string pair. Note that newly-set trace attributes are only guaranteed to propagate to *future* children of the given `Span`. See the diagram below. **(py: `set_trace_attribute`, go: `SetTraceAttribute`)**
@@ -99,7 +97,8 @@ The `Span` interface must have the following capabilities:
 
 The `Tracer` interface must have the following capabilities:
 
-- Start a `Span` that has no parent, commonly referred to as a *root* `Span` **(py: `start_trace`, go: `StartTrace`)**
+- Start a new child `Span` with a given operation name. With no parent, commonly referred to as a *root* `Span`, no trace attributes exist. With a parent any trace attributes must be passed through into the created child. **(py: `create_span`, go: `CreateSpan`)**
+- Finish an (already-started) `Span`.  Finish should be the last call made to any span instance, to do otherwise leads to undefined behavior. The parent span is returned. **(py: `finish`, go: `Finish`)**
 - Provide some form of access to a SpanPropagator (whether that's through embedding, inheritance, containment, or something else is per-platform)
 
 ### SpanPropagator
--- a/use-cases.md
+++ b/use-cases.md
@@ -10,7 +10,7 @@ This page aims to illustrate common use cases that developers who instrument the
 
 ```python
     def top_level_function():
-        span1 = tracer.start_trace('top_level_function')
+        span1 = tracer.create_span('top_level_function')
         try:
             . . . # business logic
         finally:
@@ -21,13 +21,13 @@ As a follow-up, suppose that as part of the business logic above we call another
 
 ```python
     def function2():
-        span2 = get_current_span().start_child('function2') \
+        span2 = tracer.start_child('function2', get_current_span()) \
             if get_current_span() else None
         try:
             . . . # business logic
         finally:
             if span2:
-                span2.finish()
+                tracer.finish(span2)
 ```
 
 We assume that, for whatever reason, the developer does not want to start a new trace in this function if one hasn't been started by the caller already, so we account for `get_current_span` potentially returning `None`.
@@ -46,7 +46,7 @@ When a server wants to trace execution of a request, it generally needs to go th
 
   1. Attempt to join an existing trace given a Span that's been propagated alongside the incoming request (in case the trace has already been started by the client), or create a new trace if no such propagated Span could be found.
   1. Store the newly created Span in some _request context_ that is propagated throughout the application, either by application code, or by the RPC framework.
-  1. Finally, close the Span using `span.finish()` when the server has finished processing the request.
+  1. Finally, close the Span using `tracer.finish()` when the server has finished processing the request.
 
 #### Joining to a Trace from an Incoming Request
 
@@ -113,7 +113,7 @@ In explicit propagation techniques the application code is structured to pass ar
 
     func BusinessFunction2(ctx context.Context, arg1...) {
         span := SpanFromContext(ctx)
-        span.StartChild(...)
+        tracer.create_span(..., span)
         ...
     }
 ```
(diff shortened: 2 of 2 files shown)


yurishkuro:
I don't have super strong objections to this change, but I would like to weigh pros and cons. Here's a typical Go instrumentation of in-process functions:

```go

def func1(ctx context.Context, args...) {
    childSpan := tracing.spanFromContext(ctx).startChild()
    defer childSpan.finish()

    childCtx := tracing.contextWithSpan(ctx, childSpan)
    func2(childCtx, args....)
}
```
There is no need to have access to a Tracer. The only time a Tracer is required is in the server endpoints and client call sites, both of which presumably deal with some Server and Client classes instantiated at the application startup time and thus having a reference to the Tracer, that they need to perform (de)(en)coding of the span context. But the rest of the app functions do not require a tracer.

spec.md
```diff
@@ -99,7 +97,8 @@ The `Span` interface must have the following capabilities:
 
 The `Tracer` interface must have the following capabilities:
 
-- Start a `Span` that has no parent, commonly referred to as a *root* `Span` **(py: `start_trace`, go: `StartTrace`)**
+- Start a new child `Span` with a given operation name. With no parent, commonly referred to as a *root* `Span`, no trace attributes exist. With a parent any trace attributes must be passed through into the created child. **(py: `create_span`, go: `CreateSpan`)**
```
bensigelman: 1) I am ambivalent but definitely see the value of a consolidated `create_span` call. With the caveat that Go's non-support for overloading will possibly result in a small constellation of related methods with a common prefix.

...

dkuebric: I think this may not have the flexibility we want without some flags to the method, or related methods as @bensigelman says.

More discussion in https://github.com/opentracing/opentracing-python/pull/13 but I think we must consider these 3 use cases as mentioned by @yurishkuro :

- eg. app server entry point -- start traces on incoming requests, or continue traces if context propagated
- eg. internal service entry point -- only continue traces if context propagated (no partial traces of this subsystem)
- eg. span within a process -- only continue traces from in-process parent span

My vote would be to have a `start_trace` which signals entry to a particular process etc, `create_span` which is used within a process.  This would cover eg. both the sampling and non-sampling cases.  Ideally, `start_trace` would take an optional parent span ID--in some languages, this would be augmented by `join_trace` as previously proposed, or perhaps the instrumenter can simply use `create_span` to affec this.

adriancole:
I'm not going to have time to debate this, so I'll just leave this food for thought:

Span.create_child is far more impactful that span.finish on the api. Ex. closing a span implies a lot less than creating children. I'm discussing below the design @michaelsembwever proposed except not moving span.finish

If span.finish implies reporting, vs just stopping the clock, span.finish explicitly starts the acts of reporting, or span.finish is observable and the reporting system reacts to that. IOTW, its internal structure becomes fun and interesting, but might need more care.

If finish doesn't imply reporting, then something inside the tracing system can just look for finished spans and report them. IOTW, its internal state would be pretty boring and likely not lead to FUD.

In summary, if span.finish() just means "stops the clock and is a terminal event", but doesn't directly imply reporting... then span.finish() is boring, can stay there (IMHO). We might want to say that in an implementation note some place that is quite obvious.

spec.md
michaelsembwever: @dkuebric If i've understood you correctly then these things can be done already with the existing API without needing more than the create_span method.

`create_span` covers the later two scenarios (create_trace is not needed). 

`join_trace` is just about obtaining the span from the context. You don't need to join the trace, when you've called `span = tracer.getSpanExtractor(request.class).extract(request)` then this span is already part of the trace and you're free to start using it.

The benefit of this approach is that it removes the notion of a Trace object completely from the API and the schema the user needs to load into their mind to understand how to use that API. That is the trace is just something invisible in the context that the user is never concerned about.

dkuebric: @michaelsembwever how would I implement a case where I would like to start a new Span _only if_ a previously existing trace context (eg. upstream span) exists?  Is that something that the instrumenter must now put explicitly into their logic?

The reason I ask about this case is that I think it is a common one for sampling-based tracers: you would not want to trigger a sampling decision at every Span, typically only ones at entry points into services.  If the only way to create a Span is the method `create_span` on `Tracer`, the `Tracer` implementation doesn't have much information about whether it ought to consider sampling or not.  This would demand the instrumenter put guards around many calls to `create_span`.

As an example, if `create_span` would potentially engage sampling logic in any instance where a parent Span is not provided:
```
def my_service_handler(req):
    span = create_span()
    do_stuff(span)
    span.finish()

def do_stuff(span)
    if (span)
        span2 = create_span(span)
    # xxx
    if (span)
        span2.finish()
```

As instrumentation within services for things like RPC clients, DB clients, etc grows, the case exhibited in `do_stuff` becomes much more common than the other case.

Therefore I think this is worth supporting by having two span creation methods: one which signals to the tracer that a unit of sample-able work is beginning, and one that does not.  Tracers might choose to implement the two exactly the same under the hood, but I suspect many will treat them differently.  X-Trace-based approaches benefit greatly from this today and I believe Zipkin/Dapper models do as well.

dkuebric: One other benefit of having `start_trace` handle both start + join but be separate from `create_span`: this may be important for a tracer to understand throughput in a system.  If the instrumenter is manually guarding trace points, the Tracer will not be able to know true throughput.

michaelsembwever: > Therefore I think this is worth supporting by having two span creation methods: one which signals to the tracer that a unit of sample-able work is beginning, and one that does not.

I definitely wouldn't advocate for such a burden upon the end-user.

I see your point in that `tracer.create_span(optional<span>)` fails to distinguish between a null argument that triggers a sampling logic decision and a null argument that prevents sampling.

> how would I implement a case where I would like to start a new Span only if a previously existing trace context (eg. upstream span) exists?

It was always my understanding that this is the configured sampling logic supplied to the tracer within the current process.

For example something along the lines of 
   
    def process_initialize
        sampling = Sampling.NO_NEW_TRACES
        tracer = Tracer.create(sampling)

I wouldn't think that where code is instrumented needs to be pollute with such decision making.

dkuebric: If I understand your last proposal right, it fails to allow a case where traces are started at "edge" `create_span` calls, but not started at "interior" calls, because it is a global property of the tracer and the two categories are not distinguished.  So it still requires instrumentors to do if-guards in all their subsequent instrumentation logic.  Having a separate `start_trace` method will allow this distinction to exist without boilerplate guards on every subsequent `create_span` call (which would also hide throughput counts from any interested tracer implementations, as I noted above).

michaelsembwever: @dkuebric 

> If I understand your last proposal right, it fails to allow a case where traces are started at "edge" create_span calls, but not started at "interior" calls, because it is a global property of the tracer and the two categories are not distinguished. 

No.

> So it still requires instrumentors to do if-guards in all their subsequent instrumentation logic.

No.

Given

def process_initialize
    sampling = Sampling.NO_NEW_TRACES
    tracer = Tracer.create(sampling)

it means that anywhere (interior or edge) in that process where `tracer.create_span(null)` is called a new trace is not created. Because in lei of an existing trace being present the tracer for this process is setup with a sampling decision of always false.

This avoids the need for if-guards through the code.

dkuebric: Sorry, I was unclear: in the case that you do want to start traces at the edge, you would still have this problem.  Imagine a tomcat web server which receives requests from end-users of the application.  I want to start traces when they enter tomcat, only at the HTTP server level, not be sampling for starting new traces in, say, JDBC instrumentation spans.  You would not be able to use `NO_NEW_TRACES` at a `Tracer` level, because this would preclude sampling at the edge and also the interior.

michaelsembwever: @dkuebric 

When in a tomcat process would you be creating spans off one sampling logic at the controller level but creating spans (with no parent span) off a different sampling logic?

I can't get my head around that being a valid use-case, and in the worse and unlikely situation one could just construct a separate tracer in the jdbc layer.

dkuebric: Per discussion on hangout, consolidation makes sense.  Tracers wishing to do sampling will propagate a no-trace context item to ensure that sampling occurs only at certain instrumentation points.

bensigelman:
cross-ref: https://github.com/opentracing/opentracing.github.io/issues/44#issuecomment-179654948

yurishkuro:
Go (https://github.com/opentracing/opentracing-go/pull/48) and Python (https://github.com/opentracing/opentracing-python/pull/17) address moving `start` methods to the tracer, but per broad agreement keeping `finish` on the span.

Suggest closing this in 48hrs.

yurishkuro:
@michaelsembwever are you ok to close this?

michaelsembwever:
> @michaelsembwever are you ok to close this?

Yes, it looks good. It is a positive step: a reduction in overall API surface area (collapses (potentially) duplicate methods in Tracer and Span).

PR #47, opened 2016-02-02
Remove Span.get_trace_attribute

michaelsembwever:
Remove Span.get_trace_attribute
 - simply initial api (just one way of doing things)
 - avoid Heisenberg tracing

References:
  - https://github.com/opentracing/opentracing-java/pull/7#discussion_r50498700
  - https://github.com/opentracing/opentracing.github.io/pull/47

### DIFF
--- a/use-cases.md
+++ b/use-cases.md
@@ -122,7 +122,7 @@ The downside of explicit context propagation is that it leaks what could be cons
 
 ### Tracing Client Calls
 
-When an application acts as an RPC client, it is expected to start a new tracing Span before making an outgoing request, and propagate the new Span along with that request. The following example shows how it can be done for an HTTP request. 
+When an application acts as an RPC client, it is expected to start a new tracing Span before making an outgoing request, and propagate the new Span along with that request. The following example shows how it can be done for an HTTP request.
 
 ```python
     def traced_request(request, operation, http_client):
@@ -145,7 +145,7 @@ When an application acts as an RPC client, it is expected to start a new tracing
             for key, value in h_attr.iteritems():
                 request.add_header(key, value)
 
-        # define a callback where we can finish the span 
+        # define a callback where we can finish the span
         def on_done(future):
             if future.exception():
                 span.log_event_with_payload('exception', exception)
@@ -180,7 +180,8 @@ The client and server examples above propagated the Span/Trace over the wire, in
     span.set_trace_attribute('auth-token', '.....')
 
     # server side (one or more levels down from the client)
-    token = span.get_trace_attribute('auth-token')
+    h_ctx, h_attr = tracer.propagate_span_as_text(span)
+    token = h_attr.get_trace_attribute('auth-token')
 ```
 
 ### Logging Events
@@ -190,7 +191,7 @@ We have already used `log_event_with_payload` in the client Span use case. Event
 ```python
 
     span = get_current_span()
-    span.log_event('cache-miss') 
+    span.log_event('cache-miss')
 ```
 
 The tracer automatically records a timestamp of the event, in contrast with tags that apply to the entire Span. It is also possible to associate an externally provided timestamp with the event, e.g. see [Log (Go)](https://github.com/opentracing/opentracing-go/blob/ca5c92cf/span.go#L53).
@@ -204,13 +205,13 @@ There are scenarios when it is impractical to incorporate an OpenTracing compati
 `XXX: we don't have an API for this anymore.`
 
 > Most tracing systems apply sampling to minimize the amount of trace data sent to the system.  Sometimes developers want to have a way to ensure that a particular trace is going to be recorded (sampled) by the tracing system, e.g. by including a special parameter in the HTTP request, like `debug=true`. The OpenTracing API does not have any insight into sampling techniques used by the implementation, so there is no explicit API to force it. However, the implementations are advised to recognized the `debug` Trace Attribute and take measures to record the Span. In order to pass this attribute to tracing systems that rely on pre-trace sampling, the following approach can be used:
-> 
+>
 > ```python
-> 
+>
 >     if request.get('debug'):
 >         trace_context = tracer.new_root_trace_context()
 >         trace_context.set_attribute('debug', True)
 >         span = tracer.start_span_with_context(
->             operation_name=operation, 
+>             operation_name=operation,
 >             trace_context=trace_context)
 > ```
--- a/spec.md
+++ b/spec.md
@@ -76,7 +76,6 @@ The `Span` interface must have the following capabilities:
 - Set a key:value tag on the `Span`. The key must be a `string`, and the value must be either a `string`, a `boolean`, or a numeric type. Behavior for other value types is undefined. If multiple values are set to the same key (i.e., in multiple calls), implementation behavior is also undefined. **(py: `set_tag`, go: `SetTag`)**
 - Add a new log event to the `Span`, accepting an event name `string` and an optional structured payload argument. If specified, the payload argument may be of any type and arbitrary size, though implementations are not required to retain all payload arguments (or even all parts of all payload arguments). **(py: `log_event, log_event_with_payload`, go: `LogEvent, LogEventWithPayload`)**
 - Set a trace attribute, which is a simple string:string pair. Note that newly-set trace attributes are only guaranteed to propagate to *future* children of the given `Span`. See the diagram below. **(py: `set_trace_attribute`, go: `SetTraceAttribute`)**
-- Get a trace attribute by key. **(py: `get_trace_attribute`, go: `TraceAttribute`)**
 
 ```
         [Span A]


use-cases.md
```diff
@@ -180,7 +180,8 @@ The client and server examples above propagated the Span/Trace over the wire, in
     span.set_trace_attribute('auth-token', '.....')
 
     # server side (one or more levels down from the client)
-    token = span.get_trace_attribute('auth-token')
+    h_ctx, h_attr = tracer.propagate_span_as_text(span)
+    token = h_attr.get_trace_attribute('auth-token')
```
yurishkuro: this does not remove Heisenberg effect, since the end app still can get the attribute, it just makes the API harder to use.

yurishkuro: Also, this won’t work given current semantics, because the OT tracer has no obligation to return the key/value map with any specific keys. For example, it can merge all trace attributes into a single base64 string and return a single key in the map. Or, more likely, it can prepend a certain prefix to all attribute keys so that the map keys are not mixed up with any other random HTTP headers, as well as to allow the tracer on the receiving side to read only the headers that represent trace attributes.

michaelsembwever:
For reference sake _Heisenberg Tracing_ refers to:  

When tracing goes from only instrumentation to having additional features (that the application becomes dependant on) then; 

The act of observation/instrumentation itself changes what is observed.

Now there's no such thing as pure observation-only code. Even logging interferes (synchronous logging with disk IO and latency, asynchronous with memory). And there's an argument that because the DCP mechanism is happening behind the scenes that we're already on that slippery slope and might as well slide down it.

By removing span.get_trace_attribute we're letting Span remain more of a clean instrumentational API. The functionality is still there but moved into a more flexible, and more DCP, part of the API.

Against the current SpanPropagator API the functionality might get lost, but will definitely remain with https://github.com/opentracing/opentracing.github.io/pull/45 in place.

yurishkuro:
@michaelsembwever please see my second comment above, https://github.com/opentracing/opentracing.github.io/pull/47#discussion_r51580425. We cannot rely on the name in the kv-map encoding to retrieve the attribute by the name end-user knows.

michaelsembwever:
@yurishkuro let's continue this once https://github.com/opentracing/opentracing.github.io/pull/45 is sorted. I believe this is ample ways of doing it, but it'll be more clear afterwards.

A quick example is one can use the vanilla map extractor

    tracer.getSpanInjector(Map.class).inject(span, map)
    map.get("auth-token")

But more to he point, I think we'll see clearer ways of doing even that once the Injector/Extractor discussion evolves.

yurishkuro:
agree, let's wait, but fwiw, with our tracer implementing OT the only way it would work is

```
tracer.getSpanInjector(Map.class).inject(span, map)
map.get("uberctx-auth-token")
```
because the tracer prepends `uberctx-` to all attribute names when exposing as a map, and the end user cannot rely on that.

Also, as a side note, the distributed context propagation was widely viewed as a positive feature by people at Zipkin workshop, I don't recall anyone objecting to it. In implementations like grpc, the `context` allows user code to read back the attributes.  So why, in OpenTracing, should we try so hard to hide it?

gRPC examples:
 * passing metadata: http://www.hward.com/golang-grpc-context-client-server/
 * passing trace ID as metadata: https://github.com/harlow/go-micro-services/commit/221a67bca4309aa12b69cd57ca6c1e9522c09cfc

adriancole:
I think span is the wrong place to add general purpose propagated attributes. I'll repeat a comment I've made many times, especially as DCP came up for the third time at our third workshop.

The workshop discussed a decoupled context (tag set), baggage, an independent api. It does not require routing through a tracer api to get to baggage. Ex. my scheduler, or metrics api does not need to hold a reference to the tracer api to get the same context. This is what is generally sought out (and in fact implemented in tons of frameworks!)

It is very important to understand a vote for distributed context propagation is not necessarily a vote for Span/Tracer.setAttribute(arbitrary baggage). The fact that DCP is stuffed inside the Tracer api is the polarizing topic, not the act of propagating tags. The day we finally decouple this, is the day we can stop losing time in cyclic discussion.

dkuebric:
@adriancole I agree, I think it's conceptually much better -- however, in a practical sense I'm not sure exactly how to place DCP in context with OT interface.  I thought that perhaps OT could extend a DCP API with a TraceContext object, so I prototyped it over the weekend.

Unfortunately, my results ended up being less satisfactory than I hoped.  I discovered this in the process of converting all the use-cases.md examples (and also thinking about what examples are not covered there).

In the spirit of showing reasoning behind things as you mentioned on Monday--and also hoping for a better solution from someone else--here's the PR in case anyone is interested: https://github.com/dkuebric/opentracing.github.io/pull/1

bensigelman:
@adriancole I hear you (in fact, I will point out that my initial forays here separated the DCP stuff as part of TraceContext)... the question in my mind is how to "make the simple things simple" while still supporting the cleanest DCP-friendly API we can muster. I guess I would encourage anyone and everyone to try and solve this via a diff to the core use-cases (like @dkuebric did). In my experience it was (too) challenging to both cleanly separate DCP and keep vanilla span creation and propagation "concept-lean".

bensigelman:
PS: this is an excellent thing to discuss at our inaugural hangout in a few hours!

adriancole:
Would love to join, but my flight plans seem to be clash quite consistency
with these catchups :( just wanted to send some thoughts before I became
AWOL in transit.. And on that note boarding

dkuebric:
@adriancole safe flight!  Here's the notes from @bensigelman on the hangout: https://github.com/opentracing/opentracing.github.io/issues/44#issuecomment-179654948

For the relevant part, discussion on Attributes: they are important enough that OT wants them, but context isolation not important enough for OT to make extra work for instrumentation that would be required by a clean separation of context and span.

(I would have been happy for a clean API that both separates the two and also does not require extra work in propagating multiple objects.  Unfortunately I was not able to produce one and in the process of trying to do so convinced myself it is quite a challenge.)

It will still be quite easy to implement a context-only tracer, which uses the same instrumentation (for context propagation purposes only) and no-ops Spans.  The semantic issue faced is that Span will be the interface to the Attribute data, which isn't a very obvious API as you point out above.

yurishkuro:
I think we have reached the agreement on the video call to keep read/write methods for trace attributes.

Suggest to close this in 48hrs.

michaelsembwever:
Sorry to bring this up late but…

Why not have just the `get_tag` method instead of the `get_trace_attribute` method?

Since the `get_tag` is going to return trace attributes (if trace attributes are implemented).

yurishkuro:
I am not sure what you mean, tags and attributes are different sets of data. We do not expose read function for tags in the API, they are write-only, reported to storage out of band, and tracers are not required to guarantee that all tags will even be preserved.

michaelsembwever:
> I am not sure what you mean, tags and attributes are different sets of data. We do not expose read function for tags in the API, they are write-only, reported to storage out of band, and tracers are not required to guarantee that all tags will even be preserved.

That is, has the decision been made that the two sets are kept separate.
Or… could a `get_tag` call return a trace attribute? 

That is can trace attributes be exposed via tags. This would make sense because these are the things are reported to storage. But it wouldn't make sense because of the different value types permitted.

My questioning was as much to ask for the reference to the discussion/rationale to any decision made on this front. " We do not …" doesn't cut it.

yurishkuro:
I don't have specific pointers to discussions, as this topic was touched upon in various other related discussions. I can try to summarize:
 * Currently, a Span has `setTag` and `setTraceAttribute` methods. They are separate because the semantics of tags an attributes are very different, as documented in the docs site.
 * So far nobody has requested _read_ access to the tags. In fact, no other piece of data on the Span has _read_ API aside from trace attributes. I've seen implementations of several different tracers, none of them expose span data back to the application.
 * Trace attributes have read/write API by design, because that is their raison d'être, to allow services to propagate metadata to downstream calls.

michaelsembwever:
> I can try to summarize
thanks.

bensigelman:
@yurishkuro @michaelsembwever 

+1 on Yuri's summary. I will concede that "trace attributes" never really rolled of the tongue (and still doesn't). I had a nice VC with Rodrigo Fonseca and Jonathan Mace (author of the excellent PivotTracing paper among other things) and I started to like the term "baggage" since real-life baggage is something you "carry with you", much as these key:value pairs are carried with the trace. So I am open to something like SetBaggage/GetBaggage; it's a non-standard name, but it's also a non-standard idea and would maybe help to get users to actual RTFComments (a particularly good thing in this case).

I would like to keep them totally disjoint from "span tags", though, especially since they are "beta" compared with the rest of the OT API.

yurishkuro:
I am not opposed to "baggage".

dkuebric:
I like "baggage" as well.  "Trace attribute" is probably one of the more confusing terminology issues right now.  This would make it better differentiated from tags.

PR #48, opened 2016-02-02
Simplify Span's logging methods to just "log"

michaelsembwever:
Simplify Span's logging methods to just "log"

Reference: https://github.com/opentracing/opentracing-java/pull/9

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -73,7 +73,7 @@ The `Span` interface must have the following capabilities:
 
 - Finish the (already-started) `Span`.  Finish should be the last call made to any span instance, and to do otherwise leads to undefined behavior. **(py: `finish`, go: `Finish`)**
 - Set a key:value tag on the `Span`. The key must be a `string`, and the value must be either a `string`, a `boolean`, or a numeric type. Behavior for other value types is undefined. If multiple values are set to the same key (i.e., in multiple calls), implementation behavior is also undefined. **(py: `set_tag`, go: `SetTag`)**
-- Add a new log event to the `Span`, accepting an event name `string` and an optional structured payload argument. If specified, the payload argument may be of any type and arbitrary size, though implementations are not required to retain all payload arguments (or even all parts of all payload arguments). **(py: `log_event, log_event_with_payload`, go: `LogEvent, LogEventWithPayload`)**
+- Add a new log event to the `Span`, accepting an event name `string` and an optional structured payload argument. If specified, the payload argument may be of any type and arbitrary size, though implementations are not required to retain all payload arguments (or even all parts of all payload arguments). An optional timestamp can be used to specify a past timestamp. **(py: `log`, go: `Log`)**
 - Set a trace attribute, which is a simple string:string pair. Note that newly-set trace attributes are only guaranteed to propagate to *future* children of the given `Span`. See the diagram below. **(py: `set_trace_attribute`, go: `SetTraceAttribute`)**
 - Get a trace attribute by key. **(py: `get_trace_attribute`, go: `TraceAttribute`)**
 
--- a/use-cases.md
+++ b/use-cases.md
@@ -163,7 +163,7 @@ The downside of explicit context propagation is that it leaks what could be cons
 
 ### Tracing Client Calls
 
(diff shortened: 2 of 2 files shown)


use-cases.md
```diff
@@ -158,7 +158,7 @@ When an application acts as an RPC client, it is expected to start a new tracing
             future.add_done_callback(on_done)
             return future
         except Exception e:
-            span.log_event_with_payload('exception', e)
+            span.log('exception', e)
```
yurishkuro: how would it look in Go, where there's no method overloading or default param values?

michaelsembwever: With the use of Optional parameters there needs only be one `log` method.

spec.md
```diff
@@ -74,7 +74,7 @@ The `Span` interface must have the following capabilities:
...
 - Set a key:value tag on the `Span`. The key must be a `string`, and the value must be either a `string`, a `boolean`, or a numeric type. Behavior for other value types is undefined. If multiple values are set to the same key (i.e., in multiple calls), implementation behavior is also undefined. **(py: `set_tag`, go: `SetTag`)**
-- Add a new log event to the `Span`, accepting an event name `string` and an optional structured payload argument. If specified, the payload argument may be of any type and arbitrary size, though implementations are not required to retain all payload arguments (or even all parts of all payload arguments). **(py: `log_event, log_event_with_payload`, go: `LogEvent, LogEventWithPayload`)**
+- Add a new log event to the `Span`, accepting an event name `string` and an optional structured payload argument. If specified, the payload argument may be of any type and arbitrary size, though implementations are not required to retain all payload arguments (or even all parts of all payload arguments). An optional timestamp can be used to specify a past timestamp. **(py: `log`, go: `Log`)**
```
bensigelman: @michaelsembwever there is a reasonable case to be made for consolidation around `log` in python (and indeed the examples in `use-cases.md` are pythonic)... as @yurishkuro points out, Go does not make this pretty or easy... We do presently have a `Log(data LogData)` method, but that's just the plumbing API (as you would put it!).

So if you want to point to `log`/`Log` here, that's ok with me. But if you actually want to remove all other signatures from the various implementations, I'd consider that a regression for Go.

michaelsembwever: I'm open to some languages, like go, that don't support overloading to expand to multiple methods.

bensigelman: Sounds good... well, this PR is fine with me as long as you and @yurishkuro can agree about Python.

yurishkuro: > I'm open to some languages, like go, that don't support overloading to expand to multiple method names.

I am ok with that too, but that's where we run into naming conflicts, since if we do `Log(msg)` and `LogWithPayload(msg, any)` in Go, then what do we do with the former catch-all `Log(LogData)`?

yurishkuro: @bensigelman are you thinking `LogWithTimestamp(ts, msg, payload)`?

bensigelman: @yurishkuro I was thinking that the docs here would literally be pointing to `Log(data LogData)`... it's the "superset" / base case of logging and thus the best thing to point to in the spec.

yurishkuro: @bensigelman I agree, but that breaks Mick's proposal, as what do we do with `Log(str)` in this case? Before it had a unique name `LogEvent`

bensigelman: I was thinking that the "pseudocode" would correspond to the python API since it's actually python code. :)

Someday we can bother hooking up use-case examples in *all* officially supported platforms (ideally with a page-wide selector for example language), at which point the Go examples would indeed use `LogEvent` or similar. But I like the idea of having the pointer at this line reference `Log` since it's the superset.

PS: we could also get rid of the pointers entirely in the semantic spec... though I thought they were helpful to ground things in reality.

adriancole:
If feature creep is watched carefully, I've no problem with the verb log.

bensigelman:
(seems like this has enough consensus to merge, yes?)

dkuebric:
:+1:

bensigelman:
I rebased this and resolved some minor conflicts, keeping the spirit of this PR intact (I hope/think!). I'll merge per the long-standing LGTM.

PR #53, opened 2016-02-09
use kramdown and update some misc things along the way

bensigelman:
@yurishkuro per https://github.com/opentracing/opentracing.github.io/issues/49, the site mostly "works". The only meaningful regression is that code blocks don't do syntax highlighting. If you read to the bottom at https://github.com/jekyll/jekyll/issues/2709, one does not get the sense that there will be a fix anytime soon. And it seems like we'll have to migrate over, so why not now? (??) Or something.

The rest of the changes are just minor things I noticed while perusing the various pages looking for formatting glitches.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -8,13 +8,12 @@ pages_list:
   "Contributors": '/contributors'
 
 # Dependencies
-markdown:         redcarpet
+markdown:         kramdown
 highlighter:      pygments
 
-redcarpet:
-  extensions: 
-    - autolink
-    - with_toc_data
+kramdown:
+  input: GFM
+  hard_wrap: false
 
 gems: 
   - jekyll-paginate
--- a/apis.md
+++ b/apis.md
@@ -2,11 +2,13 @@
 layout: page
 title: Per-Platform APIs
 ---
-OpenTracing APIs are available for the following platforms:
+OpenTracing APIs are in progress for the following platforms:
 
-* Golang - https://github.com/opentracing/opentracing-go
-* Python - https://github.com/opentracing/opentracing-python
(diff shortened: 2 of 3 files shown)


yurishkuro:
It would be extremely surprising if GitHub forced people to upgrade and lose syntax highlighting. This doc http://kramdown.gettalong.org/syntax.html#code-blocks says krmdown uses a different syntax for code blocks:

    ~~~ ruby
    def what?
      42
    end
    ~~~

renders as 

~~~ ruby
def what?
  42
end
~~~

otherwise the change lgtm

bensigelman:
Well, the bad news is that the syntax you cited doesn't work. I saw [the same page you presumably saw](http://jekyllrb.com/docs/templates/#code-snippet-highlighting), but to no avail.

The good news is that this idiom *does* work:

```
{% highlight python %}
def top_level_function():
    span1 = tracer.start_trace('top_level_function')                                                                   
    try:
        . . . # business logic                                                                                         
    finally:                                                                                                           
        span1.finish()                                                                                                 
{% endhighlight %}
```

So that's what I've switched things to. I don't like some of the formatting aesthetically, but it's livable and I'm out of time for this change. :)

bensigelman:
Dammit:

```
You are attempting to use the 'pygments' highlighter, which is currently unsupported on GitHub Pages. Your site will use 'rouge' for highlighting instead. To suppress this warning, change the 'highlighter' value to 'rouge' in your '_config.yml'. For more information, see https://help.github.com/articles/page-build-failed-config-file-error/#fixing-highlighting-errors.
```

I'll see what I can do with `rouge`... Should have read the Issue more carefully.

PR #55, opened 2016-02-09
Remove margins around code blocks to avoid line wrapping

yurishkuro:


yurishkuro:
@bensigelman does this work for you?  I worked for me locally

screen shot 2016-02-09 at 11 18 57 am

but does not work on .io site (after rigorous refreshing and even different browser)

screen shot 2016-02-09 at 11 20 08 am

public/css/syntax.css
```diff
@@ -59,7 +59,7 @@
 .highlight .vg { color: #033 } /* Name.Variable.Global */
 .highlight .vi { color: #033 } /* Name.Variable.Instance */
 .highlight .il { color: #f60 } /* Literal.Number.Integer.Long */
-
+figure .highlight { margin-start: 0px; margin-end: 0px; }
```
bensigelman: It took me 15 minutes to "figure" (har har har) this out, but this should read `figure.highlight` (no space)... I fixed it in master.

I have no idea why it worked locally, btw :-/

In related news, I tried and failed to get overflow text to scroll within the `<pre>` block but couldn't get it to work and gave up. I'm quite certain that it's possible, but whatever.

yurishkuro: image

Issue #58, opened 2016-02-17
clearly explain why JoinTrace requires an operation name


bensigelman:
Per https://gitter.im/opentracing/public?at=56c3ebf3c26982646215b5c7

@adriancole

bensigelman:
From http://opentracing.io/spec/#inject-and-join

> Join examines the carrier and tries to extract identifying information needed by the new Span instance it’s trying to create. **Join also accepts an operation name for that Span (since it’s typically not included within the carrier).**

adriancole:
SGTM

PR #61, opened 2016-02-17
Rough, rough draft of data semantics guidelines for discussion

dkuebric:
The OT API is becoming well-defined, but there is still not much documentation on a normative way to use it.  There's been recent discussion on the mailing list about both [HTTP span tags](https://groups.google.com/forum/#!topic/distributed-tracing/dnzo02hO8KY) as well as [span name conventions](https://groups.google.com/forum/#!topic/distributed-tracing/Cw-mE03rkgY).  There's documentation of this being developed [in examples](https://github.com/uber-common/opentracing-python-instrumentation/blob/master/opentracing_instrumentation/client_hooks/sqlalchemy.py) [for instrumentation](https://github.com/opentracing/opentracing-python/pull/17/files#r53209504), but it seems time to bring the discussion (and documentation) to a central thread.

One big concern is that there's a tradeoff between the specificity/normativeness of this guide and the ease of adoption for users: the more optional guidelines provided, the more work it may seem to an instrumentor.  For this reason, I've tried to be clear that the sections around tags and event payloads are optional and "best effort".

There's also a potential for pretty widely-varying opinions on each of these specs.  To limit initial discussion scope, I've started by only documenting the tag/log schemes we have already used in various examples or codebases.  There are a number of other semantic specs that would be very useful to add (around database/cache usage in particular), which I can see us adding as a follow up.


data-semantics.md
```diff
@@ -0,0 +1,106 @@
...
+
+
+## Spans
+
+
+### Span Naming
+
+Spans carry tags, logs, and attributes, but they also have a top-level **operation name**. This should be a low-cardinality string value representing the type of work being done in the span. Typically, a process, framework, library, or module name is a good guideline.
+
+For instance, a request being handled by a Flask-based application might be instrumented using a WSGI middleware. In this situation, the first span operation name might be `WSGI`. If the application subsequently makes a database query, the span name might be `SQL`.
```
dkuebric: This is a slightly different take than currently exists in many implementations, which suggest a HTTP method name or similar for the span.  Because we are recommending these type of details be added via tags, I suggest a more general service name approach.

data-semantics.md
```diff
@@ -0,0 +1,106 @@
...
+
+
+### Span Structure
+
+TODO
```
dkuebric: I think this may be well-covered in the [Semantic Specification](opentracing.io/spec/) doc today, so I haven't duplicated the contents.  However, I can also see a use-case for adding a discussion about span structure in various concurrency scenarios.

data-semantics.md
```diff
@@ -0,0 +1,106 @@
...
+### Exception
+
+* Event: `exception`
+* Payload:
+    - `type` - string
+        - Low-cardinality class of exception
+    - `val` - string
+        - Details associated with particular instance of this exception
+    - `tb` - string
+        - Stack trace provided at exception time
```
adriancole: Sorry, what does tb stand for?

data-semantics.md
```diff
@@ -0,0 +1,106 @@
...
+
+## Semantic Log Use-Cases
+
+Some events are point-in-time and do not apply to an entire span; these are reported via logged events.  For instance, an error or exception may be reported via a `log` call.
+
+
+### Exception
+
+* Event: `exception`
+* Payload:
```
adriancole: First use of payload which I can relate to ;)

adriancole:
Thanks for starting this. Just curious, from a practical perspective, do you expect appneta to use these? Or are these general suggestions that might work for others.

data-semantics.md
dkuebric: Traceback, which is a python-ism for a stack trace.  I don't know that these are optimal--I stole the ones used in the python implementation, and they're pretty terse.

How about these instead: `type`, `message`, and `backtrace`?

dkuebric:
@adriancole AppNeta relies on a somewhat rich set of semantic guidelines--we have about 2 dozen specs for different span and event types.  I excerpted what we do for HTTP server spans in a distributed-tracing discussion thread: https://groups.google.com/d/msg/distributed-tracing/dnzo02hO8KY/54vdbgxABwAJ

I'd love to see a comparable level of detail emerge here because I think it's almost necessary in order for the instrumentation to be truly portable across tracers.  Tracers benefit their users by adding more capabilities, but if any capabilities are dependent on particular tag names, and the names are non-standard, that breaks down some of the portability advantages of the API.

data-semantics.md
```diff
@@ -0,0 +1,106 @@
...
+The following tags are recommended for instrumentors who are trying to represent a particular type of data. Tag names follow a general structure of namespacing; the values are determined purely on a tag-by-tag basis.
+
+Some tags mentioned below may contain values of significant size. Handling of such values is implementation-specific: it is the responsibility of the Tracer to honor, drop, or truncate these tags as appropriate. However, care should be exercised on the part of the instrumentor, as even the generation or passing of such values to the Tracer may create undesirable overhead for the application.
+
+It is not required that all specified tags be used, even if one is used.
+
+
+### HTTP Server Tags
+
+These tags are recommended for spans marking entry into a HTTP-based service.
```
yurishkuro: I would like to defined these in `ext.tags` and announce in the documentation that people should use the constants rather than string names below. Maybe even change documentation to refer to constants (a bit difficult due to different languages).

data-semantics.md
yurishkuro: I think this puts a different meaning on the operation name. What you're suggesting sounds more like the layer/component, not the operation name. For example, if an app is using Redis client, then the span calling to Redis would have operation_name = get or put, but component_name=Redis_Client

adriancole: I share Yuri's opinion. Ex I see component name as operation name a
degraded case. Ex people complain in HTTP when the name is as vague as the
method name. This is admittedly biased towards how UIs present it.

dkuebric: I think operation is valuable, but it seems to miss a step in the pyramid--the step that defines what exactly is doing the operating.  There's certain things we are expecting we can provide automatically via traces, like hostname, which will help localize where work is being done.  Component feels as though it ought to be one, though it's not one that I would expect most tracers to be able to fetch automatically as hostname, so we would expect users to provide in the same way they provide op.

@yurishkuro Is `component_name` a standardized thing?  I don't see it in any of the codebases, but I do think it makes sense as a label for this.  Potentially just `component` as well.  One issue with your Redis example is I could see the same operation names being used for memcache, or HTTP--there's a bit of ambiguity if you skip the `component` element.

@adriancole I agree that more precision must be provided.  There is some UI bias inherent in terms of how we're thinking about this (from myself included!).

At root, the issue is that a single attribute of a span (`operation_name`) has been elevated to the top level, and everything else is subjugated as tags.  I expect in many cases the `operation_name` ought to be supplemented by further tags.  OT will benefit from being as precise as possible about what is a good type of value for this top-level attribute vs the other tags.

So I'm fine with trying to keep `operation_name` as some sort of operation name, and adding a concept of `component` separately, but conceptually it feels clearer to have the broader-strokes `component` as our elevated attribute.  Either way, we will want to tighten this up.  Could be as simple as a set of examples and some guidelines (eg. I think "low cardinality" is a good one)?

data-semantics.md
dkuebric: Agreed--some tracers might want to use the shortest possible representation to report, vs full string.  I think as long as the consts look generally similar across languages, we could use one as an example here for each, along with a section explaining the abstraction of the underlying label?

data-semantics.md
yurishkuro: @dkuebric I agree, we should at least make `component` into a standard tag. I already had to deal with it when generating Zipkin's "local spans", which do expect such component name. I also agree that often the importance of operation vs. component varies. @adriancole already questioned why we require op-name on join (here https://github.com/opentracing/opentracing.github.io/issues/58). Conceptually they can both be provided as tags, or both be elevated to proper concept (e.g. we have `span.setOperationName`). 

There is another aspect to it. I think "component" refers to a "local component" or "layer" within the application. E.g. a business service would never have a component called "redis", but it may have "redis_client". The Redis server identifies itself to the tracer at tracer creation time via impl-specific API, which may include both service name ("trips_cache") and service kind ("redis").

bensigelman: I am ok with the idea of standardizing around a component as a package... I am not enthusiastic about component as a process name: it makes more sense to make that an attribute of the Tracer impl since there will not be more than one process per Tracer impl.

data-semantics.md
bensigelman: (I like those suggestions, @dkuebric)

adriancole:
meta-comment.. this looks like the first time that the "payload" is
actually required to be used. (otherwise there's just "exception"). We can
expect those who have payload-related formatting concerns to have concerns
here.. for example, I suppose a tracing system who export logs as a
single-string would need to map a pattern across (some of) these fields.

Ex in the case of a string-cat based tracer impl, I suppose it would need
be ok with logging being passing a map and then grabbing the fields and
passing them to a string-cat joiner. cc @bogdandrutu

data-semantics.md
dkuebric: Summarizing the above, I think we're into `component` as a thing, and "package" feels like a decent abstraction to capture it.  The last decision is how formalized it becomes vs `operation_name`.

It seems as though there's 4 categories of option:

1. Demote `operation_name` and strongly suggest both it and `component`
  - I don't think we want to remove all mandatory span labeling, so this is out.

2. Elevate `component` to equivalent (`span.setComponentName` and inclusion in constructors)
  - This is tempting to me.  Would it be considered too onerous?

3. Replace `operation_name` with `component` and strongly suggest `operation_name` as a tag
  - This is the direction AppNeta went, though it's the opposite of Zipkin I believe?  Potentially controversial.

4. Strongly suggest `component` as a tag
  - I suspect this will end up with weaker compliance than we want.

My vote is 2 or 3.  Thoughts?

data-semantics.md
dkuebric: Fixed that up.

data-semantics.md
bensigelman: I vote for (4). For what it's worth, I would argue strenuously against (1). I would also argue strenuously against (3): it makes sense for monkey-patched instrumentation, but OT is designed for explicit instrumentation and the operation name is going to be more descriptive semantically. (2) is conceivable but feels burdensome... though maybe I'm just too accustomed to the status quo?

I appreciate why `component` can be useful, but (a) information about the package can be captured automatically by the tracer impl, and (b) it is easier to add `setComponentName` later than to take it away. So, yes, my vote is (4) with the option to pursue (2) in the future once we see how useful it is to have both.

yurishkuro: @bensigelman 

> information about the package can be captured automatically by the tracer impl

can you elaborate? Typically there's going to be a single tracer per process, but many components/layers per process, like marshaling layer, caching, sql client, any business component the application might nominate.

bensigelman: @yurishkuro @dkuebric 

> > information about the package can be captured automatically by the tracer impl
>
> can you elaborate?

I'm talking about grabbing fully-qualified package names from the stack, etc, etc... logging packages have various ways of doing this. Of course the available options vary by language.

yurishkuro: > I'm talking about grabbing fully-qualified package names from the stack, etc, etc.

Hmm, I wouldn't put my faith in the tracer always getting it right via this method.

@dkuebric in my company getting just the RPCs instrumented and traced across the whole architecture is a major challenge, and RPC spans do not benefit from component as much as from the operation name. I am not sure if `component` even makes sense in RPC case - would it be something like the specific type of server/client framework, e.g. Dropwizard or CurlHttpClient?

So at this point I would vote for (4).

yurishkuro:
@dkuebric maybe split the "operation name" discussion and merge the rest? The common tags are a useful thing to have publicly available (we also need to define `ext` constants).

data-semantics.md
```diff
@@ -0,0 +1,106 @@
...
+    - Query string parameters of the request being handled.
+    - Note that the preceding `?` character is not expected.
+    - Examples:
+        - `key=val&key2=val2`
+* `http.method` - string
+    - HTTP method of the request being handled.
+    - Case-insensitive
+    - Examples:
+        - `GET`, `POST`, `HEAD`
+* `http.status` - integer
```
yurishkuro: maybe `status_code`?  Cf. https://github.com/openzipkin/zipkin/blob/master/zipkin-thrift/src/main/thrift/com/twitter/zipkin/zipkinCore.thrift#L157

dkuebric:
@yurishkuro sounds good, I've forked the `component` discussion to #75 .  I'll get the other items (status_code, ext tag references, flesh out span structure section reference) sorted so we can merge the less controversial stuff.

dkuebric:
@yurishkuro still not sure how to represent the ext tags -- possibly add a list of links to ext tag packages in the `opentracing-*` repos to the top section?  Also reference one or two example implementations for each tag next to string, i.e.:

* `peer.hostname` (py: `ext.tags.PEER_HOSTNAME`, go: `ext.PeerHostname.Set()`)

yurishkuro:
@dkuebric definitely useful to link to the source files in each repo, as they are mostly self-explanatory. The per-language example for each individual tag would be fantastic, but too tedious to maintain, I would probably limit that to just one tag, to show the usage pattern.

I think long term (https://github.com/opentracing/opentracing.github.io/issues/76) it would be great to have a yaml file in the docs repo that defines each tag (name, description, value type), then symlink the docs repo into each language repo as a git submodule, and write simple code generators. That way we would achieve uniform descriptions of the tags, make sure that each std tag is defined in each langauge, etc. The actual html doc can also be generated from the same yaml file.  All the generators can probably be written once using the same template framework.

dkuebric:
OK, I think we're converging towards something that might be mergeable.  I've incorporated current feedback and squashed all the intermediary edits.  Opened a fresh PR for further discussion: #82

Issue #70, opened 2016-02-24
Make the Tracer inject/join (i.e., propagate) directly rather than through an additional delegation layer


bensigelman:
At present, the [OT semantic spec](http://opentracing.io/spec/) has both a `Tracer.injector(format)` method/concept and an `Injector.inject()` method/concept. Strictly speaking this is not necessary, though OT implementations could (or could not, whatever) choose to create helper libraries that would have similar abstractions. To review, at present the injection code path looks (in pseudo-python) like:

```
some_carrier = SomeCarrier()
my_tracer.injector(Format.SOME_FORMAT).inject(span, some_carrier)
```

It could alternatively look like...

```
some_carrier = SomeCarrier()
my_tracer.inject(span, Format.SOME_FORMAT, some_carrier)
```

... or, in languages with stronger types, even just ...

```
someCarrier := NewSomeCarrier()  // each carrier gets its own type
myTracer.Inject(span, someCarrier)
```

(i.e., no `format` argument as a type specifier)

----------------

The above also came up here: https://github.com/opentracing/opentracing-java/pull/10/files#r53574474 and here: https://github.com/opentracing/opentracing-go/pull/48#issuecomment-181047181; @tschottdorf said:

> Maybe I'm not reading this right, but are you basically suggesting going towards opentracing/opentracing-go#48 (comment)? I.e. (this is copy-pasta from there):
> 
> ```
> type Tracer interface {
>   // Join attempts to obtain a Span from the given interface.
>   Join (interface{}) (Span, error)
> }
> 
> type Span interface {
>   // Inject attempts to store the Span's information in the given interface.
>   Propagate(interface{}) error
> }
> 
> t := something.NewTracer()
> sp := t.StartTrace("foo")
> var req http.Request
> // This will add the trace information to the request's headers.
> if err := t.Propagate(&req); err != nil {
>   return err
> }
> 
> // Receiving end uses the Tracer to join a new Span onto the one
> // encoded in the Request.
> if sp, err := t.Join(&req); err != nil {
>   // ...
> }
> ```

---------------------

I'm not sure if I'm to the point of advocating for the current approach or the simplified version that has fewer types/concepts (as there are some advantages to each), but I wanted to make sure that we were intentional as a group about which path we take.

Curious to hear thoughts from those who've had opinions along these lines thus far: @yurishkuro @michaelsembwever

yurishkuro:
I am in favor of the syntax that accepts the format:
```
[...]

PR #72, opened 2016-03-01
Remove the Injector/Extractor layer in the spec

bensigelman:
Per https://github.com/opentracing/opentracing.github.io/issues/70

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -49,7 +49,7 @@ Every Span has zero or more **Logs**, each of which being a timestamped event na
 
 Every Span may also have zero or more key/value **Tags**, which do not have timestamps and simply annotate the spans.
 
-Spans may be **Injected** into and **Extracted** from objects that are used for inter-process communication (e.g., HTTP headers). In this way, Spans may propagate across process boundaries along with sufficient information to join up with the Trace in some remote process.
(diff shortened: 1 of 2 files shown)


spec.md
```diff
@@ -126,22 +120,19 @@ The "text" and "binary" designations refer to two flavors of encoding:
...
 
 To make the above more concrete, consider the following sequence:
 
-1. A client process has a `Span` instance and is about to make an RPC over a home-grown HTTP protocol
-1. That client process calls some `Tracer.Injector(...)` method, passing a format identifier for a string map
-1. With an `Injector` instance in hand, the client calls `injector.InjectSpan(...)`, passing the active `Span` instance and a string map carrier in as parameters
+1. A *client* process has a `Span` instance and is about to make an RPC over a home-grown HTTP protocol
+1. That client process calls some `Tracer.Inject(...)` method, passing the active `Span` instance, a format identifier for a string map, and a string map carrier as parameters
 1. The newly-populated (opaque) string map is encoded into the homegrown HTTP protocol (e.g., as headers) by the application code
 1. Now in the server process, the application code grabs the opaque string map from the homegrown HTTP protocol
```
yurishkuro: the term "opaque map" is quite vague here, could mean different things, i.e. it could be that the user knows it's a map but does not know what it contains, or that a user doesn't even know it's a map (because it is "encoded into headers").

Not sure if it can be expressed more clearly without overwhelming the reader with details better left for elsewhere.

yurishkuro:
LGTM

spec.md
bensigelman: I made some edits to clarify this a bit...

PR #80, opened 2016-03-18
[GH-27] Document global tracer and no-op tracer requirements.

### LINKED ISSUE #27: Should GlobalTracer be a no-op tracer by default?
As we've decided to go with having a Global tracer implementation for ease-of-use (see prior discussion at opentracing/opentracing-go#10), a thought has crossed my mind:

1. I'd really like to be able to submit PRs to open-source projects that may not know what tracing is, and in such cases they might be more inclined to like using globals rather than modifying their existing API.
2. They wouldn't want existing users of their API to have to do OpenTracing setup with a Noop tracer implementation, just to get the same behavior as previously.
3. This leaves me with an idea: should GlobalTracer default to being the no-op tracer implementation? Then libraries could make calls to opentracing safely and it'd be no-op by default unless the user of such a library opted-in to installing their tracing implementation as the global.

Thoughts?

beberlei:
Closes #27

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -146,3 +146,9 @@ To make the above more concrete, consider the following sequence:
 1. In the absence of data corruption or other errors, the *server* now has a `Span` instance that belongs to the same trace as the one in the client
 
 More concrete examples can be found among the [OpenTracing use cases](/use-cases).
+
+### No-op Tracer and Global Tracer
+
+OpenTracing implementations must provide a No-Op Tracer as part of their interface. The no-op tracer and spans do neither support tracing nor baggage propagation. The No-op implementation must be written in a way that it does not crash or otherwise produce side-effects/errors when being used by developers. Most notably it must provide a No-op span implementation as well, so that developers can rely on a span instance to be returned from the Tracer Span Starting and Joining functions.
+
+OpenTracing implementations should provide support for configuring (Go: `InitGlobalTracer()`, py: `opentracing.tracer = myTracer`) and retrieving (Go: `GlobalTracer()`, py: `opentracing.tracer`) a global tracer instance if this is possible from the platform perspective. This should be implemented as a static registry or with a global variable. The default global tracer must be the No-op tracer.


adriancole:
Since tracer requires plumbing ops like join and inject, they should
probably be noted.

Ex simplest case could be that join returns no-op span (more obvious) and
inject on a no-op span adds no bytes or strings to the carrier (more
subtle).

More subtle unless inject already says it writes nothing if the span is
no-op or otherwise unwritable.

spec.md
```diff
@@ -146,3 +146,9 @@ To make the above more concrete, consider the following sequence:
 1. In the absence of data corruption or other errors, the *server* now has a `Span` instance that belongs to the same trace as the one in the client
 
 More concrete examples can be found among the [OpenTracing use cases](/use-cases).
+
+### No-op Tracer and Global Tracer
+
+OpenTracing implementations must provide a No-Op Tracer as part of their interface. The no-op tracer and spans do neither support tracing nor baggage propagation. The No-op implementation must be written in a way that it does not crash or otherwise produce side-effects/errors when being used by developers. Most notably it must provide a No-op span implementation as well, so that developers can rely on a span instance to be returned from the Tracer Span Starting and Joining functions.
```
bensigelman: grammar nit: `s/do neither support/support neither/`

bensigelman: it's permissible for `Tracer.Join` to return no span if there was no trace to join... so maybe we only need to specify that Span-starting needs to return something that won't crash the process.

spec.md
```diff
@@ -146,3 +146,9 @@ To make the above more concrete, consider the following sequence:
...
+
+OpenTracing implementations should provide support for configuring (Go: `InitGlobalTracer()`, py: `opentracing.tracer = myTracer`) and retrieving (Go: `GlobalTracer()`, py: `opentracing.tracer`) a global tracer instance if this is possible from the platform perspective. This should be implemented as a static registry or with a global variable. The default global tracer must be the No-op tracer.
```
bensigelman: SG!

bensigelman:
LGTM aside from the nits. Thanks!

spec.md
yurishkuro: I think noop implementations should be (and already are) provided by the OpenTracing API libraries themselves, not by pluggable implementations.

bensigelman: definitely agree with @yurishkuro -- by "OpenTracing implementations", I understood this to refer to the various platforms...

bensigelman:
(I'm going to adopt this branch and get it in)

bensigelman:
(thank you, @beberlei... I merged in your change plus some copy-edits in https://github.com/opentracing/opentracing.github.io/pull/88)

Issue #76, opened 2016-03-06
Consolidate common tags definition and auto-generate language bindings


yurishkuro:
Pulled out of https://github.com/opentracing/opentracing.github.io/pull/61

I think long term it would be great to have a yaml file in the docs repo that defines each tag (name, description, value type), then symlink the docs repo into each language repo as a git submodule, and write simple code generators. That way we would achieve uniform descriptions of the tags, make sure that each std tag is defined in each langauge, etc. The actual html doc can also be generated from the same yaml file.  All the generators can probably be written once using the same template framework.

bensigelman:
+1 -- great idea.

bensigelman:
(c.f. #92)

bensigelman:
... continued at https://github.com/opentracing/specification/issues/10

Issue #75, opened 2016-03-06
Top-level tags: "operation_name" and/or "component"?


dkuebric:
Per #61 , there's a question about the recommendation for contents of the `operation_name` tag as well as importance/role/value of a `component` tag.  In order to close out other parts of that PR, I've opened this at @yurishkuro suggestion to continue the discussion.

Recap:

The only "required" span tag currently is, `operation_name`, which is valuable and currently a cornerstone of many UI displays.  However, it seems to miss a step in the semantic pyramid--the step that defines what exactly is doing the operating. There's certain things we are expecting we can provide automatically via traces, like hostname, which will help localize where work is being done. `component` feels as though it ought to be one, though it's not one that I would expect most tracers to be able to fetch automatically as hostname.  So if we think this is valuable, we would expect instrumentors to provide in the same way they provide op.

Additionally, `operation_name` has been elevated to a special status; I think for a combination of historical and practical reasons.  I'm fine with trying to keep `operation_name` as some sort of operation name, and adding a concept of component separately, but conceptually it feels clearer to have the broader-strokes component as our elevated attribute.

At any rate, we will want to tighten on recommendations for `operation_name`.

It seems as though there's 4 categories of option:

1. Demote `operation_name` and strongly suggest both it and `component`
  - I don't think we want to remove all mandatory span labeling, so this is out.
2. Elevate `component` to equivalent (`span.setComponentName` and inclusion in constructors)
  - This is tempting to me.  Would it be considered too onerous?
3. Replace `operation_name` with `component` and strongly suggest `operation_name` as a tag
  - This is the direction AppNeta went, though it's the opposite of Zipkin I believe?  Potentially controversial.
4. Strongly suggest `component` as a tag
  - I suspect this will end up with weaker compliance than we want.

My vote is 2 or 3.

@bensigelman had expressed interest in pursuing 4 and a strong opposition to 1/3.
@yurishkuro had expressed interest in pursuing 4.

dkuebric:
Examples of potential `component` vs `operation_name` distinction for discussion:
* HTTP client call
  - `component`: `httplib`
  - `operation_name`: `GET | POST | ...`
  - Other attributes like URL, status code, etc specified in extended tags
* Memcache client call
  - `component`: `redis-py`
  - `operation_name`: `GET | HGETALL | APPEND | ...`
* Template engine
  - `component`: `jinja2`
  - `operation_name`: `profile.html | account.html | ...`

[...]

PR #81, opened 2016-03-18
[GH-75][GH-76] Document "component" tag and all other standardized tags in spec

beberlei:
See #75, but also documented all other tags in the spec (that is not done yet) as per #76

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -120,6 +120,20 @@ Note that `Inject` and `Join` are not entirely symmetrical interfaces since `Spa
 
 A note about **"format identifiers"** per the `Inject` and `Join` methods above: the nature of the "format identifier" may vary from platform to platform, but in all cases they should be drawn from a global namespace. New formats must not *require* changes to the core OpenTracing platform APIs, though those core platform APIs must define a few basic/general formats (like string maps and binary blobs). For example, if the maintainer of EsotericRPCFramework wanted to define an EsotericRPCFramework inject/join format, she or he must be able to do so without sending a PR to OpenTracing maintainers (though of course OpenTracing implementations are not required to support the EsotericRPCFramework format). There is [an end-to-end injector and extractor example below](#propagation-example) to make this more concrete.
 
+### Standardized Tags
+
+There is a number of standardized tags that should be used accross platforms and tracing implementations:
+
+- `peer.service` records the service name of the peer
+- `peer.host` records the host name of the peer
+- `peer.ipv4` records IP v4 host address of the peer
+- `peer.ipv6` records IP v6 host address of the peer
+- `peer.port`  records port number of the peer
+- `span.kind` hints at relationship between spans, e.g. values for client `c` and server `s`.
+- `component` optionally records the component that the current span belongs to.
+
+Implementations should provide constants for these tag names.
+
 ### Required Inject/Join carrier formats
 
 At a minimum, all platforms require OpenTracing implementations to support two carrier formats: the "text map" format and the "binary" format.


adriancole:
These look reasonable, especially inside the tracer itself as it is
implicitly only one side of an operation. I like the labels.

There is some trouble in systems where a span includes both the client and
server spans (ex zipkin v1), a query requires context as they would repeat.
Ex two component tags in the same span, and queries usually imply single
value.

For example, would a zipkin span that includes a client and server be kind
client or kind server? Which component wins? I'm not saying zipkin like
systems don't already have this problem, just noting impact. Ex In zipkin
we have local_component, which notes this is for a single-host span. That's
what local means.. That the span doesn't include two sides of an rpc.

I'm not saying don't publish these, just noting impact that might steer if
or how to note usage.

dkuebric:
@adriancole good point.  Not being intimately familiar with the implementation, would it be possible for the Zipkin collector to take advantage of these keys even as it reduces to a single span?

For instance, would you see a Zipkin OT tracer implementation reporting the same span twice, once from each side of client/server?  If so, when the collection tier stitches them together, it could use redundant/conflicting data like `span.kind` and `component` to further validate the spans and choose to preserve or index whichever the Zipkin model prefers today, or even rename tags on one side to disambiguate them if that is preferred.

yurishkuro:
@dkuebric this is some missing (imo) functionality in Zipkin query and UI for preserving the ownership of data elements in those combined spans, even though it is available in the raw data. I do think there is value in each side reporting its own `component`, e.g. when client uses one http lib and the server uses another. All the Zipkin UI needs to do is to annotate the tags with the service name that sent them, even if they are merged into a single "UI span"

adriancole:
I should have guessed this thread would be hijacked as zipkin ui feature
request. I'm not going to indulge this further, as history has taught me
such an act will just soak up 2 hours rehashing things.

adriancole:
I'm sure I'm digging a hole, but to close the side-bar https://github.com/openzipkin/zipkin/issues/963 is zipkin issue on presenting shared spans. There's a display column to disambiguate annotations: log events, but not tags (binary annotations).

Events vary on time and host, basically, and eventhough the underlying zipkin datastructure for tags includes host, it isn't displayed.



Why? a couple reasons. First, it hasn't become important enough for anyone to actually raise a pull request. Also, there has been a majority of zipkin contributors that think the underlying tag datastructure should be a string->string per span. The complexity of the "binary annotation" object is actually the opposite, if you listen to others. The UI isn't wrong, the data structure should be simpler! But guess what.. no-one has changed that either.

back to my original suggestion. 

> I'm not saying don't publish these, just noting impact that might steer if
or how to note usage.

That's all I meant. Just consider an impact that may not be obvious. There are many tracing systems not participating on this thread, for example, and they may have a similar issue.

bensigelman:
@adriancole I don’t have nearly enough context on the open Zipkin issue or – especially – the history behind it (nor am I asking for such context), but FWIW I read @yurishkuro’s [comment above](https://github.com/opentracing/opentracing.github.io/pull/81#issuecomment-198803584) as an attempt to offer an on-topic opinion:

> I do think there is value in each side reporting its own `component`

... in the context of a real tracing system (Zipkin). In particular, I did not read Yuri's comment as a feature request about Zipkin. Though I agree that this PR is not the place to make such requests.

.......

As for the `component` itself: I am supportive of a `component` tag even in the RPC case; and even if the client and server components are not identical. The worst case scenario is that a tracing system has to drop the client- or server-side tag: IMO that's preferable to not-having the tag in the first place (as "component" is well-defined within the host process).

yurishkuro:
@adriancole apologies if my comment sounded dismissive of your concern.

I do not believe that Zipkin v1 model is fundamentally incapable of capturing identical tag names submitted separately by server and client tracers. Cassandra backend captures that data properly and makes it available in the raw data, making the disambiguation a UI issue. I could have clarified my comment by mentioning Cassandra backend. But by the same token, you could have also qualified your comment that the issue is only with _some_ Zipkin backends, which for some reasons _choose to ignore the data dimension they receive from upstream_.

Even for those Zipkin backends that literally merge spans at the storage level and lose association of tags, they can pre-process the spans at ingestion, as @dkuebric [mentioned above](https://github.com/opentracing/opentracing.github.io/pull/81#issuecomment-198776761). If a tracer follows OT recommendation, it would also emit "span kind" tag, so that collector (which still sees spans pre-merge) can replace `component` with `client-component` or `server-component`.

I agree, it is worth mentioning this approach on the "Implementing OT" page, but I don't think it should affect the data specification or how people instrument their apps/libs with OT.

yurishkuro:
btw, I think this PR is superseded by a larger change in #82

beberlei:
Closing in favour of #82

PR #82, opened 2016-03-19
Add data semantics

dkuebric:
This is a cleaned-up version of #61 to continue the path towards merge, after discussion there and in #75.

I haven't incorporated the auto-generation suggested in #76 yet but that would be a possible next step.

One question in my mind is whether this document should get listed in the top-level nav?  For now, I've left it out and referenced from the `Log` and `Span` sections of `spec.md`.


data-semantics.md
```diff
@@ -0,0 +1,117 @@
...
+
+As an example, consider the common case of a HTTP-based application server. The URL of an incoming request that the application is handling is often useful for diagnostics, as well as the HTTP verb and the resultant status code. An instrumentor could choose to report the URL in a tag named `URL`, or perhaps named `http.url`--either would be valid from the pure API perspective. But if the Tracer wishes to add intelligence, such as indexing on the URL value or sampling proactively for requests to a particular endpoint, it must know where to look for relevant data. In short, when tag names and other instrumentor-provided values are used consistently, the tracers on the other side of the API can employ more intelligence.
+
+The guidelines contained describe a common ground on which instrumentors and tracer authors can build beyond pure data collection. Adherence to the guidelines is optional but highly recommended for instrumetors.
```
yurishkuro: "contained here", and typo in the last word "instrumetors"

data-semantics.md
```diff
@@ -0,0 +1,117 @@
...
+
+Some tags mentioned below may contain values of significant size. Handling of such values is implementation-specific: it is the responsibility of the Tracer to honor, drop, or truncate these tags as appropriate. However, care should be exercised on the part of the instrumentor, as even the generation or passing of such values to the Tracer may create undesirable overhead for the application.
+
+It is not required that all suggested tags be used, even if one is used.
+
+### Component Identification
+
+For any span, it can be useful to specify the type of component being instrumented. This is particularly recommended for instrumentation provided for libraries or modules, where end-users may haev a mix of custom and library-provided instrumentation.
```
yurishkuro: typo "may *haev*a mix"

data-semantics.md
```diff
@@ -0,0 +1,117 @@
...
+
+### HTTP Server Tags
+
+These tags are recommended for spans marking entry into a HTTP-based service.
+
+* `http.url` - string
+    - URL of the request being handled in this segment of the trace, in [standard URI format](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier), excepting query string parameters.
+    - Protocol optional
+    - Examples:
+        - `https://domain.net/path/to?resource=here`
```
yurishkuro: this includes query string

data-semantics.md
```diff
@@ -0,0 +1,117 @@
...
+        - `domain.net/path/to/resource`
+        - `http://user:pass@domain.org:8888`
+* `http.query` - string
```
yurishkuro: @dkuebric can you elaborate why, in your experience, it is valuable to have instrumentation separate query string from the rest of the URL?  it seems like something the implementation could easily do with `split("?")`

data-semantics.md
```diff
@@ -0,0 +1,117 @@
...
+    - Query string parameters of the request being handled.
+    - Note that the preceding `?` character is not expected.
+    - Examples:
+        - `key=val&key2=val2`
+* `http.method` - string
+    - HTTP method of the request being handled.
+    - Case-insensitive
+    - Examples:
+        - `GET`, `POST`, `HEAD`
+* `http.status_code` - integer
```
yurishkuro: is integer always available?  I seem to recall running into string values in some http APIs.

data-semantics.md
```diff
@@ -0,0 +1,117 @@
...
+
+## Semantic Log Use-Cases
+
+Some events are point-in-time and do not apply to an entire span; these are reported via logged events.  For instance, an error or exception may be reported via a `log` call.
+
+
+### Exception
+
+* Event: `exception`
+* Payload:
```
yurishkuro: this probably needs language examples - which structure is used to encode the payload?

yurishkuro:
Overall LGTM, a few minor typos, and a couple questions.

I think it makes sense to include in the top level menu, under Semantic Spec link.

data-semantics.md
```diff
@@ -0,0 +1,117 @@
...
+The guidelines contained describe a common ground on which instrumentors and tracer authors can build beyond pure data collection. Adherence to the guidelines is optional but highly recommended for instrumetors.
+
+
+
+## Spans
+
+
+### Span Naming
+
+Spans carry tags, logs, and attributes, but they also have a top-level **operation name**. This should be a low-cardinality string value representing the type of work being done in the span. Typically, this is the local type of work being done. Examples include `GET` or `POST` for a HTTP span, `SELECT` or `INSERT` for a SQL span, and so on.
```
bensigelman: s/attributes/baggage/

bensigelman: I see the need for `GET`/`SELECT`/etc as operation names when nothing else is available, but in an ideal world the operation name would have more semantic meaning... e.g., in an RPC-based system it would be the name of the RPC (thrift/grpc/whatever) method. Maybe that would be a better example?

data-semantics.md
```diff
@@ -0,0 +1,117 @@
...
+Secondarily, an optional **component** tag can be provided to scope the operation name. Typically, a process, framework, library, or module name is a good guideline for the component tag.
+
+### Span Structure
+
+Span structure is also important: what do spans represent, and what relationship do they have towards their parent span?  This is covered in the [Semantic Specification](/spec).
+
+
+## Span Tag Use-Cases
+
+The following tags are recommended for instrumentors who are trying to represent a particular type of data. Tag names follow a general structure of namespacing; the values are determined purely on a tag-by-tag basis.
```
bensigelman: I am probably being dense, but what do you mean by "the values are determined purely on a tag-by-tag basis"? (I.e., as opposed to what?)

data-semantics.md
```diff
@@ -0,0 +1,117 @@
...
+    - Examples:
+        - `https://domain.net/path/to?resource=here`
+        - `domain.net/path/to/resource`
+        - `http://user:pass@domain.org:8888`
+* `http.query` - string
+    - Query string parameters of the request being handled.
+    - Note that the preceding `?` character is not expected.
+    - Examples:
+        - `key=val&key2=val2`
+* `http.method` - string
```
bensigelman: I like the idea of `http.method` as a tag, but how does this fit in with the guidance about HTTP verbs as operation names? (C.f. my comment that they might not be the best example of operation nameS)

data-semantics.md
```diff
@@ -0,0 +1,117 @@
...
+    - HTTP method of the request being handled.
+    - Case-insensitive
+    - Examples:
+        - `GET`, `POST`, `HEAD`
+* `http.status_code` - integer
+    - HTTP status code to be returned with HTTP response.
+    - Examples:
+        - `200`, `503`
+* `span.kind` - string
+    - Value of `s` should be used to indicate that this is a server side span (see "Peer Tags")
```
bensigelman: nit: link to `#peertags` and add that anchor below?

bensigelman: nit: `server` rather than `s`? otherwise seems like a premature optimization to me. tracer impls can compress these if they see fit.

data-semantics.md
```diff
@@ -0,0 +1,117 @@
...
+Spans carry tags, logs, and attributes, but they also have a top-level **operation name**. This should be a low-cardinality string value representing the type of work being done in the span. Typically, this is the local type of work being done. Examples include `GET` or `POST` for a HTTP span, `SELECT` or `INSERT` for a SQL span, and so on.
+
+Secondarily, an optional **component** tag can be provided to scope the operation name. Typically, a process, framework, library, or module name is a good guideline for the component tag.
+
+### Span Structure
+
+Span structure is also important: what do spans represent, and what relationship do they have towards their parent span?  This is covered in the [Semantic Specification](/spec).
+
+
+## Span Tag Use-Cases
```
bensigelman: (errors are really important, but I think we should express that as a proper method on `Span`...)

bensigelman:
Looking great! I just made a bunch of nits.

data-semantics.md
yurishkuro: I agree, that would be a better example, especially as a guideline. In practice, when implicit instrumentation is used, it's nearly impossible to come up with business-specific names, but if an RPC framework itself is instrumented, it often does know the name of the endpoint it is calling.

We had a situation where we had a thick client for a storage service, which was communicating to the server via http. The implicit instrumentation picked the default `get/post` names at the http level spans, but the explicit instrumentation in the thick client was wrapping them into better-named spans.

bensigelman:
(@dkuebric friendly ping on this)

dkuebric:
@yurishkuro @bensigelman thanks for all the thoughtful comments (and close reading).  I've taken most of your suggestions directly.  I'm hoping this is mergeable right now, as I've heard some others had changes they wanted to start merging in.  Here's my thoughts on some of the less obvious changes I made:

1. Status code.  @yurishkuro -- could conceivably either be int or string.  I don't know that one is a clear winner, but based on a quick survey of AppNeta typed instrumentation (java, c-based), int is more common format at framework/handler level.  Looked at apache, nginx, PHP, netty, and servlet interfaces.  So I'd vote we suggest int for now--seems simpler than support for both/either.
2. Query string.  @yurishkuro -- first off, thanks for catching the immediate inconsistency of the advice ;).  Second, I think that the idea to split it into 2 parts is perhaps not something to tackle at this layer--it comes from some specific requests we have gotten to not report query string data gathered from instrumentation.  This could be handled by tracer internals.  So, I've removed `http.query` as a separate tag and suggested inclusion in base URI tag.
3. Error payload.  @bensigelman -- agreed this may be best handled via a dedicated method.  To the extent that log_event is used, I've updated it to suggest a map-based type (dict, etc) per @yurishkuro suggestion.
4. `client` and `server` string.  Expanded from `c` and `s`, which I stole from zipkin.  This actually feels like a good candidate for encoding in ext tags?

data-semantics.md
bensigelman: I take this back :)

Per http://opentracing.io/2015/04/14/weekly-hangout/:

> TL;DR: we will formalize a tag name+value ("error": true, most likely) for span errors

I'm happy to make that change myself (in fact, it's why I wanted to merge this in), but we won't add a dedicated method until that's deemed necessary.

data-semantics.md
```diff
@@ -32,7 +32,7 @@ Span structure is also important: what do spans represent, and what relationship
 
 ## Span Tag Use-Cases
 
-The following tags are recommended for instrumentors who are trying to represent a particular type of data. Tag names follow a general structure of namespacing; the values are determined purely on a tag-by-tag basis.
+The following tags are recommended for instrumentors who are trying to represent a particular type of data. Tag names follow a general structure of namespacing.
 
 The recommended tags below are accompanied by `const` values included in an `ext` module for each opentracing implementation.  These `ext` values should be used in place of the strings below, as tracers may choose to use different underlying representations for these common concepts.  The symbols are similar in each implementation: ([Go](https://github.com/opentracing/opentracing-go/blob/master/ext/tags.go), [Python](https://github.com/opentracing/opentracing-python/blob/master/opentracing/ext/tags.py), etc.)
```
bensigelman: Just FYI (i.e., no need to change this PR), https://github.com/opentracing/opentracing.github.io/issues/76 seems like a very wise idea and – once realized – would be something to reference here... a canonical YAML file and codegen seems like the only sane way to go in any case.

data-semantics.md
```diff
@@ -0,0 +1,112 @@
...
+* `http.method` - string
+    - HTTP method of the request being handled.
+    - Case-insensitive
+    - Examples:
+        - `GET`, `POST`, `HEAD`
+* `http.status_code` - integer
+    - HTTP status code to be returned with HTTP response.
+    - Examples:
+        - `200`, `503`
+* `span.kind` - string
```
yurishkuro: I think this should be its own section, and higher on the list. In particular, Zipkin-compatible tracers cannot work without this tag unless they infer RPC from the use of inject/join, which is indirect at best.

yurishkuro:
LGTM

>  client and server string. Expanded from c and s, which I stole from zipkin. This actually feels like a good candidate for encoding in ext tags?

Already there, but as c/s.  We can change that, since there tags are normally only used in-memory. https://github.com/opentracing/opentracing-go/blob/master/ext/tags.go#L43

bensigelman:
(LGTM as well)

data-semantics.md
dkuebric: What would you think about placing `span.kind` under the "component identification" section?  I think it should still be referenced here, but maybe removed from the "peer tags" in favor of this other, higher visibility placement.

yurishkuro: Sounds good.

data-semantics.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Semantic Log Use-Cases
+
+Some events are point-in-time and do not apply to an entire span; these are reported via logged events.  For instance, an error or exception may be reported via a `log` call.
+
+
+### Exception
+
+* Event: `exception`
+* Payload - map
```
yurishkuro: I think `map` is too language specific (i.e. python). If this was Java/JavaScript, I would simply do `span.log("exception", exc)` and let the tracer figure out how to record it.

dkuebric: Agree that languages would benefit from some convenience methods that automatically extract features from commonly-used error/exception objects.  However, would that be implemented via a log_event overload, or via a log_error or similar?

I suppose secretly I've been assuming that payloads should are maps/arrays/objects in the way that spans are.  (An event might have an arbitrary set of data to report with it.)  If not, instrumentors are going to have to figure out what types of payloads might be valid for various tracer implementations, I guess?

yurishkuro: > However, would that be implemented via a log_event overload, or via a log_error or similar?

No, I was thinking just plain old `log` method: `span.log("exception", exc)`

> If not, instrumentors are going to have to figure out what types of payloads might be valid for various tracer implementations, I guess?

If we were shooting for maps, they I would suggest changing the API to take a map rather than `interface{}` / `Object`. But it feels that people would question this choice; after all, an instance of Exception is a natural representation of the exception, why make more work for people to break it apart into a map?

We have stayed away from defining semantics of the payload so far. It's definitely something that we need to collectively address and be more prescriptive, to clarify the contract between instrumentors and implementors.  But I would suggest a different issue/PR for that. Until then, an opaque `Object` is the least restrictive form.

bensigelman: FWIW, I have generally thought of the `payload` first and foremost as a place for the OpenTracing caller to stick "FYI"-level structured data that might help explain or contextualize the trace.

bensigelman:
@dkuebric can you merge this soon?

dkuebric:
@bensigelman Made changes per discussions of other week.  I don't have write access but I think the PR is ready, pending a quick review.

bensigelman:
@dkuebric oh dear, sorry! I'll review and likely merge shortly.

dkuebric:
👍

PR #91, opened 2016-04-26
Update broken link

ifraixedes:
The previous link didn't exist, I think that you wanted to link to integrations.


yurishkuro:
Good spot, thanks!

Issue #99, opened 2016-06-20
introduce a formal SpanContext concept at the OT layer


bensigelman:
For those who think better with code, there's an illustration of the change [in opentracing-go](https://github.com/opentracing/opentracing-go/pull/82) and a demonstration of the ramifications for implementations and calling code [in basictracer-go](https://github.com/opentracing/basictracer-go/pull/29).

The benefits of a change like this are manifold:

- Inject+Join are replaced by Inject+Extract which operate exclusively at the metadata/context layer
- Join goes away and all Spans are created by a single StartSpan API
- Multiple-parent / DAG trace structure maps cleanly to the API (with all span-to-span pointers using metadata/context)
- Span use-after-`Finish` semantics are clean and clear (it's fine to reference the metadata/context and nothing else)
- The baggage set/get API moves to the metadata/context type
- Span metadata/context can be used to identify a Span in structured log messages

(Related to https://github.com/opentracing/opentracing.github.io/issues/96 https://github.com/opentracing/opentracing.github.io/issues/95 https://github.com/opentracing/opentracing.github.io/issues/28 https://github.com/opentracing/opentracing.github.io/issues/25 https://github.com/opentracing/opentracing.github.io/issues/18)

bensigelman:
Historical note: the initial attempt at a standard OpenTracing API had something like a span metadata/context type (called `TraceContext`)... that version of things was considerably more complicated with a separate factory type to create the `TraceContext`s. There was an attempt to have span-to-span references expressed with the `TraceContext` API, and that turned out to be problematic since most tracing systems send linkage information *out-of-band* (i.e., not in the `TraceContext`) which led to complex APIs. And though the simplification from that initial attempt to what we have now was a net improvement, in retrospect I think we threw the baby out with the bathwater.

There's been plenty of discussion about the merits of this proposal on https://github.com/opentracing/opentracing-go/pull/82 and it looks very much like it will go forward.

TODO:
- [ ] get LGTM(s) on the terminology in the Go proof-of-concept PR
- [ ] write up a PR for the documentation site and make sure things look good at that layer
- [ ] merge the above, then...
- [ ] try to fast-track analogous PRs for the other OT libs (and unblock C# and PHP libs which are both in progress)

bensigelman:
Well, https://github.com/opentracing/opentracing.github.io/issues/106 is still open, but at this point all of the fully-baked OT APIs have SpanContext. (Still waiting on C++ and C#, but those are behind in other ways, too)

PR #100, opened 2016-06-25
Introduce SpanContext, s/Join/Extract/, etc

bensigelman:
Per https://github.com/opentracing/opentracing.github.io/issues/99 and https://github.com/opentracing/opentracing-go/pull/82 .

### DIFF
--- a/data-semantics.md
+++ b/data-semantics.md
@@ -27,7 +27,7 @@ Secondarily, an optional **component** tag can be provided to scope the operatio
 
 ### Span Structure
 
-Span structure is also important: what do spans represent, and what relationship do they have towards their ancestor spans?  This is covered in the [Semantic Specification](/spec).
+Span structure is also important: what do spans represent, and what relationship do they have towards their ancestors?  This is covered in the [Semantic Specification](/spec).
 
 
 ## Span Tag Use-Cases
--- a/data-semantics.md
+++ b/data-semantics.md
@@ -27,7 +27,7 @@ Secondarily, an optional **component** tag can be provided to scope the operatio
 
 ### Span Structure
 
-Span structure is also important: what do spans represent, and what relationship do they have towards their parent span?  This is covered in the [Semantic Specification](/spec).
+Span structure is also important: what do spans represent, and what relationship do they have towards their ancestor spans?  This is covered in the [Semantic Specification](/spec).
 
 
 ## Span Tag Use-Cases
--- a/propagation.md
+++ b/propagation.md
@@ -1,151 +1,147 @@
 ---
 layout: page
-title: Everything You Wanted to Know About Inject and Join but were Afraid to Ask
+title: Everything You Wanted to Know About Inject and Extract but were Afraid to Ask
 ---
 
-Programmers adding tracing support across process boundaries must understand the `Tracer.Inject(...)` and `Tracer.Join(...)` capabilities of [the OpenTracing specification](/spec). They are conceptually powerful, allowing the programmer to write *correct* and *general* cross-process propagation code **without being bound to a particular OpenTracing implementation**; that said, with great power comes great opportunity for confusion. :)
+Programmers adding tracing support across process boundaries must understand the `Tracer.Inject(...)` and `Tracer.Extract(...)` capabilities of [the OpenTracing specification](/spec). They are conceptually powerful, allowing the programmer to write *correct* and *general* cross-process propagation code **without being bound to a particular OpenTracing implementation**; that said, with great power comes great opportunity for confusion. :)
 
-This document provides a concise summary of the design and proper use of `Inject` and `Join`, regardless of the particular OpenTracing language or OpenTracing implementation.
+This document provides a concise summary of the design and proper use of `Inject` and `Extract`, regardless of the particular OpenTracing language or OpenTracing implementation.
 
 ## "The Big Picture" for explicit trace propagation
 
 The hardest thing about distributed tracing is the *distributed* part. Any tracing system needs a way of understanding the causal relationship between activities in many distinct processes, whether they be connected via formal RPC frameworks, pub-sub systems, generic message queues, ad hoc HTTP calls, best-effort UDP packets, or something else entirely.
 
-Some distributed tracing systems (e.g., [Project5](http://dl.acm.org/citation.cfm?id=945454) from 2003, or [WAP5](http://www2006.org/programme/item.php?id=2033) from 2006 or [The Mystery Machine](https://www.usenix.org/node/186168) from 2014) *infer* causal relationships across process boundaries. Of course **there is a tradeoff between the apparent convenience of these black-box inference-based approaches and the ultimate freshness and quality of the assembled traces.** Per the concern about quality, OpenTracing is an *explicit* distributed tracing instrumentation standard, and as such it is much better-aligned with approaches like [X-Trace](https://www.usenix.org/conference/nsdi-07/x-trace-pervasive-network-tracing-framework) from 2007, [Dapper](http://research.google.com/pubs/pub36356.html) from 2010, or numerous open-source tracing systems like [Zipkin](https://github.com/openzipkin) or [Appdash](https://github.com/sourcegraph/appdash) (among others).
+Some distributed tracing systems (e.g., [Project5](http://dl.acm.org/citation.cfm?id=945454) from 2003, or [WAP5](http://www2006.org/programme/item.php?id=2033) from 2006 or [The Mystery Machine](https://www.usenix.org/node/186168) from 2014) *infer* causal relationships across process boundaries. Of course **there is a tradeoff between the apparent convenience of these black-box inference-based approaches and the freshness and quality of the assembled traces.** Per the concern about quality, OpenTracing is an *explicit* distributed tracing instrumentation standard, and as such it is much better-aligned with approaches like [X-Trace](https://www.usenix.org/conference/nsdi-07/x-trace-pervasive-network-tracing-framework) from 2007, [Dapper](http://research.google.com/pubs/pub36356.html) from 2010, or numerous open-source tracing systems like [Zipkin](https://github.com/openzipkin) or [Appdash](https://github.com/sourcegraph/appdash) (among others).
 
 Let's briefly restate the "Why OpenTracing?" section of the [OpenTracing landing page](/):
 
 > By offering consistent, expressive, vendor-neutral APIs for popular platforms, OpenTracing makes it easy for developers to add (or switch) tracing implementations with an O(1) configuration change. OpenTracing also offers a lingua franca for OSS instrumentation and platform-specific tracing helper libraries.
 
-**Together, `Inject` and `Join` allow for inter-process trace propagation without tightly coupling the programmer to a particular OpenTracing implementation.** 
+**Together, `Inject` and `Extract` allow for inter-process trace propagation without tightly coupling the programmer to a particular OpenTracing implementation.** 
 
 ## Requirements for the OpenTracing propagation scheme
 
-For `Inject` and `Join` to be useful, all of the following must be true:
+For `Inject` and `Extract` to be useful, all of the following *must* be true:
 
 - Per the above, the [OpenTracing user](/use-cases/#stepping-back-who-is-opentracing-for) handling cross-process trace propagation must not need to write OpenTracing-implementation-specific code
 - OpenTracing implementations must not need special handlers for every known inter-process communication mechanism: that's far too much work, and it's not even well-defined
 - That said, the propagation mechanism should be extensible for optimizations
 
-## The basic approach: Inject, Join, and Carriers
+## The basic approach: Inject, Extract, and Carriers
 
-Any Span in a trace may be **Injected** into what OpenTracing refers to as a Carrier. A **Carrier** is an interface or data structure useful for inter-process communication (IPC); that is, the Carrier is something that "carries" the tracing state from one process to another. The OpenTracing specification includes two [required Carrier formats](#required-carriers), though [custom Carrier formats](#custom-carriers) are possible as well.
+Any SpanContext in a trace may be **Injected** into what OpenTracing refers to as a Carrier. A **Carrier** is an interface or data structure that's *useful* for inter-process communication (IPC); that is, the Carrier is something that "carries" the tracing state from one process to another. The OpenTracing specification includes two [required Carrier formats](#required-carriers), though [custom Carrier formats](#custom-carriers) are possible as well.
 
-Similarly, given a Carrier, an injected trace may be **Joined**, yielding a new Span. We use the term "Join" rather than "Extract" (or similar) because Inject/Join is not a symmetric process: an OpenTracing implementation is only expected to inject the minimal amount of information needed to constunct a (descendant) Span in the peer and join it to the overarching trace. The injected state is *not* necessarily sufficient to reconstruct the entire injected Span instance, symmetrically.
+Similarly, given a Carrier, an injected trace may be **Extracted**, yielding a SpanContext instance which is semantically identical to the one Injected into the Carrier.
 
 #### Inject pseudocode example
 
 {% highlight python %}
-span = ...
-outboundRequest = ...
+span_context = ...
+outbound_request = ...
 
 # We'll use the (builtin) TEXT_MAP carrier format. We start by
 # using an empty map as the carrier prior to the call to
 # `tracer.inject`.
 carrier = {}
-tracer.inject(span, opentracing.Format.TEXT_MAP, carrier)
+tracer.inject(span_context, opentracing.Format.TEXT_MAP, carrier)
 
 # `carrier` now contains (opaque) key:value pairs which we pass
 # along over whatever wire protocol we already use.
 for key, value in carrier:
-    outboundRequest.headers[key] = escape(value)
+    outbound_request.headers[key] = escape(value)
 {% endhighlight %}
 
-#### Join pseudocode example
+#### Extract pseudocode example
 
 {% highlight python %}
-inboundRequest = ...
+inbound_request = ...
 
 # We'll again use the (builtin) TEXT_MAP carrier format. Per the
 # TEXT_MAP documentation, we can use a map that has extraneous data
 # in it and let the OpenTracing implementation look for the subset
 # of key:value pairs it needs.
 #
-# As such, we directly use the key:value `inboundRequest.headers`
+# As such, we directly use the key:value `inbound_request.headers`
 # map as the carrier.
-carrier = inboundRequest.headers
-# The operation name we want for the server-side span:
-operationName = ...
-serverSpan = tracer.join(
-    operationName, opentracing.Format.TEXT_MAP, carrier)
+carrier = inbound_request.headers
+span_context = tracer.extract(opentracing.Format.TEXT_MAP, carrier)
+// Continue the trace given span_context. E.g.,
+//
+//     span = tracer.start_span("...", parent=span_context)
+//
 
-# (If `carrier` held trace data, `span` will now be ready to use.)
+# (If `carrier` held trace data, `span_context` will now be ready to use.)
 {% endhighlight %}
 
 #### Carriers have formats
 
 All Carriers have a format. In some OpenTracing languages, the format must be specified explicitly as a constant or string; in others, the format is inferred from the Carrier's static type information.
 
-#### Not Symmetric
-
-When a Span is injected into a Carrier, the OpenTracing implementation only encodes the minimum amount of information needed to join with the trace in a remote process. As such, **`Inject` and `Join` are not glorified serialization and deserialization methods:** the Span returned by `Join` is not the Span passed to `Inject`.
-
 <div id="required-carriers"></div>
 
-## Required Inject/Join Carrier formats
+## Required Inject/Extract Carrier formats
 
 At a minimum, all platforms require OpenTracing implementations to support two Carrier formats: the "text map" format and the "binary" format.
 
 - The *text map* Carrier format is a platform-idiomatic map from (unicode) `string` to `string`
 - The *binary* Carrier format is an opaque byte array (and presumably more compact and efficient)
 
-What the OpenTracing implementations choose to store in these Carriers is not formally defined by the OpenTracing specification, but the presumption is that they find a way to encode "tracer state" about the propagated `Span` (e.g., in Dapper this would include a `trace_id`, a `span_id`, and a bitmask representing the sampling status for the given trace) as well as any key:value Baggage items.
+What the OpenTracing implementations choose to store in these Carriers is not formally defined by the OpenTracing specification, but the presumption is that they find a way to encode "tracer state" about the propagated `SpanContext` (e.g., in Dapper this would include a `trace_id`, a `span_id`, and a bitmask representing the sampling status for the given trace) as well as any key:value Baggage items.
 
 ### Interoperability of OpenTracing implementations *across process boundaries*
 
-There is no expectation that different OpenTracing implementations `Inject` and `Join` Spans in compatible ways. Though OpenTracing is agnostic about the tracing implementation *across an entire distributed system*, for successful inter-process handoff it's essential that the processes on both sides of a propagation use the same tracing implementation.
+There is no expectation that different OpenTracing implementations `Inject` and `Extract` SpanContexts in compatible ways. Though OpenTracing is agnostic about the tracing implementation *across an entire distributed system*, for successful inter-process handoff it's essential that the processes on both sides of a propagation use the same tracing implementation.
 
 <div id="custom-carriers"></div>
 
-## Custom Inject/Join Carrier formats
+## Custom Inject/Extract Carrier formats
 
-Any propagation subsystem (an RPC library, a message queue, etc) may choose to introduce their own custom Inject/Join Carrier format; by preferring their custom format **but falling back to a required OpenTracing format as needed** they allow OpenTracing implemantations to optimize for their custom format without *needing* OpenTracing implementations to support their format.
-
+Any propagation subsystem (an RPC library, a message queue, etc) may choose to introduce their own custom Inject/Extract Carrier format; by preferring their custom format **but falling back to a required OpenTracing format as needed** they allow OpenTracing implemantations to optimize for their custom format without *needing* OpenTracing implementations to support their format.
 
 Some pseudocode will make this less abstract. Imagine that we're the author of the (sadly fictitious) **ArrrPC pirate RPC subsystem**, and we want to add OpenTracing support to our outbound RPC requests. Minus some error handling, our pseudocode might look like this:
 
 {% highlight python %}
-span = ...
-outboundRequest = ...
+span_context = ...
+outbound_request = ...
 
-# First we try our custom Carrier, the outboundRequest itself.
+# First we try our custom Carrier, the outbound_request itself.
 # If the underlying OpenTracing implementation cares to support
 # it, this call is presumably more efficient in this process
 # and over the wire. But, since this is a non-required format,
 # we must also account for the possibility that the OpenTracing
 # implementation does not support arrrpc.ARRRPC_OT_CARRIER.
 try:
-    tracer.inject(span, arrrpc.ARRRPC_OT_CARRIER, outboundRequest)
+    tracer.inject(span_context, arrrpc.ARRRPC_OT_CARRIER, outbound_request)
 
 except opentracing.UnsupportedFormatException:
     # If unsupported, fall back on a required OpenTracing format.
     carrier = {}
-    tracer.inject(span, opentracing.Format.TEXT_MAP, carrier)
+    tracer.inject(span_context, opentracing.Format.TEXT_MAP, carrier)
     # `carrier` now contains (opaque) key:value pairs which we
     # pass along over whatever wire protocol we already use.
     for key, value in carrier:
-	outboundRequest.headers[key] = escape(value)
+	outbound_request.headers[key] = escape(value)
 {% endhighlight %}
 
 <div id="format-identifiers"></div>
 
 ### More about custom Carrier formats
 
-The precise representation of the "Carrier formats" may vary from platform to platform, but in all cases they should be drawn from a global namespace. New formats must not *require* changes to the core OpenTracing platform APIs, though those core platform APIs must define a few basic/general formats (like string maps and binary blobs). For example, if the maintainer of ArrrPC RPC framework wanted to define an "ArrrPC" inject/join format, she or he must be able to do so without sending a PR to OpenTracing maintainers (though of course OpenTracing implementations are not required to support the "ArrrPC" format). There is [an end-to-end injector and extractor example below](#propagation-example) to make this more concrete.
+The precise representation of the "Carrier formats" may vary from platform to platform, but in all cases they should be drawn from a global namespace. Support for a new custom carrier format *must not necessitate* changes to the core OpenTracing platform APIs, though each OpenTracing platform API must define the required OpenTracing carrier formats (e.g., string maps and binary blobs). For example, if the maintainer of ArrrPC RPC framework wanted to define an "ArrrPC" Inject/Extract format, she or he must be able to do so without sending a PR to OpenTracing maintainers (though of course OpenTracing implementations are not required to support the "ArrrPC" format). There is [an end-to-end injector and extractor example below](#propagation-example) to make this more concrete.
 
 
 <div id="propagation-example"></div>
 
-## An end-to-end Inject and Join propagation example
+## An end-to-end Inject and Extract propagation example
 
 To make the above more concrete, consider the following sequence:
 
-1. A *client* process has a `Span` instance and is about to make an RPC over a home-grown HTTP protocol
-1. That client process calls `Tracer.Inject(...)`, passing the active `Span` instance, a format identifier for a text map, and a text map Carrier as parameters
+1. A *client* process has a `SpanContext` instance and is about to make an RPC over a home-grown HTTP protocol
+1. That client process calls `Tracer.Inject(...)`, passing the active `SpanContext` instance, a format identifier for a text map, and a text map Carrier as parameters
 1. `Inject` has populated the text map in the Carrier; the client application encodes that map within its homegrown HTTP protocol (e.g., as headers)
 1. *The HTTP request happens and the data crosses process boundaries...*
 1. Now in the server process, the application code decodes the text map from the homegrown HTTP protocol and uses it to initialize a text map Carrier
-1. The server process calls `Tracer.Join(...)`, passing in the desired operation name, a format identifier for a text map, and the Carrier from above
-1. In the absence of data corruption or other errors, the *server* now has a `Span` instance that belongs to the same trace as the one in the client
+1. The server process calls `Tracer.Extract(...)`, passing in the desired operation name, a format identifier for a text map, and the Carrier from above
+1. In the absence of data corruption or other errors, the *server* now has a `SpanContext` instance that belongs to the same trace as the one in the client
 
 Other examples can be found in the [OpenTracing use cases](/use-cases) doc.
--- a/propagation.md
+++ b/propagation.md
@@ -70,7 +70,7 @@ span_context = tracer.extract(opentracing.Format.TEXT_MAP, carrier)
 //     span = tracer.start_span("...", parent=span_context)
 //
 
-# (If `carrier` held trace data, `span_context` will now be ready to use.)
+# (If `carrier` held trace data, `span` will now be ready to use.)
 {% endhighlight %}
 
 #### Carriers have formats
--- a/propagation.md
+++ b/propagation.md
@@ -67,7 +67,7 @@ carrier = inbound_request.headers
 span_context = tracer.extract(opentracing.Format.TEXT_MAP, carrier)
 // Continue the trace given span_context. E.g.,
 //
-//     span = tracer.start_span("...", parent=span_context)
+//     span = tracer.start_span("...", belongs_to=span_context)
 //
 
 # (If `carrier` held trace data, `span` will now be ready to use.)
--- a/propagation.md
+++ b/propagation.md
@@ -65,10 +65,8 @@ inbound_request = ...
 # map as the carrier.
 carrier = inbound_request.headers
 span_context = tracer.extract(opentracing.Format.TEXT_MAP, carrier)
-// Continue the trace given span_context. E.g.,
-//
-//     span = tracer.start_span("...", child_of=span_context)
-//
+# Continue the trace given span_context. E.g.,
+span = tracer.start_span("...", child_of=span_context)
 
 # (If `carrier` held trace data, `span` will now be ready to use.)
 {% endhighlight %}
--- a/propagation.md
+++ b/propagation.md
@@ -67,7 +67,7 @@ carrier = inbound_request.headers
 span_context = tracer.extract(opentracing.Format.TEXT_MAP, carrier)
 // Continue the trace given span_context. E.g.,
 //
-//     span = tracer.start_span("...", belongs_to=span_context)
+//     span = tracer.start_span("...", child_of=span_context)
 //
 
 # (If `carrier` held trace data, `span` will now be ready to use.)
--- a/propagation.md
+++ b/propagation.md
@@ -96,7 +96,7 @@ There is no expectation that different OpenTracing implementations `Inject` and 
 
 ## Custom Inject/Extract Carrier formats
 
-Any propagation subsystem (an RPC library, a message queue, etc) may choose to introduce their own custom Inject/Extract Carrier format; by preferring their custom format **but falling back to a required OpenTracing format as needed** they allow OpenTracing implemantations to optimize for their custom format without *needing* OpenTracing implementations to support their format.
+Any propagation subsystem (an RPC library, a message queue, etc) may choose to introduce their own custom Inject/Extract Carrier format; by preferring their custom format **but falling back to a required OpenTracing format as needed** they allow OpenTracing implementations to optimize for their custom format without *needing* OpenTracing implementations to support their format.
 
 Some pseudocode will make this less abstract. Imagine that we're the author of the (sadly fictitious) **ArrrPC pirate RPC subsystem**, and we want to add OpenTracing support to our outbound RPC requests. Minus some error handling, our pseudocode might look like this:
 
--- a/spec.md
+++ b/spec.md
@@ -93,7 +93,7 @@ As is the case with Logs, if certain known tag key:values are used for common ap
 
 ## SpanContext
 
-Every Span must provide access to a **SpanContext**. The SpanContext represents Span state that must propagate to descendant Spans and across process boundaries (e.g., a `<trace_id, span_id, sampled>` tuple) and also encapsulates any **Baggage** (see below). SpanContext is used when propagating traces across process boundaries and when creating edges in the trace graph (e.g., parent-child relationships).
+Every Span must provide access to a **SpanContext**. The SpanContext represents Span state that must propagate to descendant Spans and across process boundaries (e.g., a `<trace_id, span_id, sampled>` tuple) and also encapsulates any **Baggage** (see below). SpanContext is used when propagating traces across process boundaries and when creating edges in the trace graph (e.g., parent-child relationships or other [references](#references)).
 
 ### Baggage
 
@@ -110,7 +110,7 @@ Also, Baggage keys have a restricted format: implementations may wish to use the
 
 ## Inject and Extract
 
-SpanContexts may be **Injected** into and **Extracted** from **Carrier** objects that are used for inter-process communication (e.g., HTTP headers). In this way, SpanContexts may propagate across process boundaries along with sufficient information to join up with the Trace in some remote process.
+SpanContexts may be **Injected** into and **Extracted** from **Carrier** objects that are used for inter-process communication (e.g., HTTP headers). In this way, SpanContexts may propagate across process boundaries along with sufficient information to reference Spans (and thus continue traces) from remote processes.
 
 # Platform-Independent API Semantics
 
--- a/spec.md
+++ b/spec.md
@@ -54,13 +54,24 @@ A **Span** represents a logical unit of work in the system that has a start time
 
 A Span may reference zero or more Spans that are causally related. OpenTracing presently defines two types of references: `ChildOf` and `FollowsFrom`. **Both reference types specifically model direct causal relationships between a child span and a parent span.** In the future, OpenTracing may also support reference types for spans with non-causal relationships (e.g., Spans that are batched together, Spans that are stuck in the same queue, etc).
 
-**`ChildOf` references:** A Span may be the "ChildOf" a parent Span. In a "ChildOf" reference, the parent Span depends on the child Span in some fashion (often—but not always—the parent Span cannot finish until the child Span finishes).
+**`ChildOf` references:** A Span may be the "ChildOf" a parent Span. In a "ChildOf" reference, the parent Span depends on the child Span in some capacity. All of the following would constitute ChildOf relationships:
 
-A typical timing diagram for a Span that is the ChildOf a parent Span:
+- A Span representing the server side of an RPC may be the ChildOf a Span representing the client side of that RPC
+- A Span representing a SQL insert may be the ChildOf a Span representing an ORM save method
+- Many Spans doing concurrent (perhaps distributed) work may all individually be the ChildOf a single parent Span that merges the results for all children that return within a deadline
+
+These could all be valid timing diagrams for children that "FollowFrom" a parent.
 
 ~~~
     [-Parent Span---------]
          [-Child Span----]
+
+    [-Parent Span--------------]
+         [-Child Span A----]
+          [-Child Span B----]
+        [-Child Span C----]
+         [-Child Span D---------------]
+         [-Child Span E----]
 ~~~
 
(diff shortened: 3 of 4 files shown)


spec.md
```diff
@@ -8,66 +8,96 @@ title: Semantic Specification
...
- [Span D]      +---+-----+
-               |         |
-           [Span E]  [Span F]
-                         |
-                +--------+--------+
-                |        |        |
-            [Span G] [Span H] [Span I]
+ [Span D]      +---+-------+
+               |           |
+           [Span E]    [Span F] >>> [Span G] >>> [Span H]
```
bensigelman: tried to represent "sequence predecessors" in these diagrams...

spec.md
```diff
@@ -8,66 +8,96 @@ title: Semantic Specification
...
+A **Trace** represents the potentially distributed, potentially concurrent data/execution path in a (potentially distributed, potentially concurrent) system. A Trace can be thought of as a DAG of Spans. (See the ASCII diagrams above)
 
 ## Spans
 
-A **Span** represents a logical unit of work in the system that has a start time and a duration. Spans may be nested and ordered to model parent-child and causal relationships. Each span has an **operation name**, a presumably human-readable string which concisely names the work done by the span (e.g., an RPC method name, a function name, or the name of a subtask within a larger computation).
+A **Span** represents a logical unit of work in the system that has a start time and a duration. Spans may be nested and ordered to model causal relationships. Each span has an **operation name**, a presumably human-readable string which concisely names the work done by the span (e.g., an RPC method name, a function name, or the name of a subtask or stage within a larger computation).
+
+<span id="references"></span>
+
+### Causal Span References
```
bensigelman: new concept for the docs site...

spec.md
```diff
@@ -78,51 +108,60 @@ Baggage comes with powerful _costs_ as well; since the Baggage is propagated in-
...
-
-Note that `Inject` and `Join` are not entirely symmetrical interfaces since `Span` injection is a fundamentally lossy process (most of the `Span`'s state is meant to be recorded out-of-band from the application's own communication path, whereas Inject-ed state is meant to be sent in-band along with the application data).
```
bensigelman: so glad there's no longer a need to explain this!

use-cases.md
```diff
@@ -173,13 +174,13 @@ def traced_request(request, operation, http_client):
...
         operation_name=operation,
-        parent=parent_span,
+        parent=parent_span.context(),
         tags={'http.url': request.full_url}
     )
 
     # propagate the Span via HTTP request headers
     tracer.inject(
-        span,
+        span.context(),
```
yurishkuro: I think in Python we'd want context to be exposed as a `@property` (or even a stored attribute, but `@property` can make it read only), so the syntax would be `span.context`, not a function call.

spec.md
```diff
@@ -78,51 +108,60 @@ Baggage comes with powerful _costs_ as well; since the Baggage is propagated in-
...
+- **Add a new log event** to the `Span`, accepting an event name `string` and an optional structured payload argument. If specified, the payload argument may be of any type and arbitrary size, though implementations are not required to retain all payload arguments (or even all parts of all payload arguments). An optional timestamp can be used to specify a past timestamp. **(py: `log`, go: `Log`)**
+
+## The `SpanContext` Interface
+
+The `SpanContext` interface must have the following capabilities. The user acquires a reference to a `SpanContext` via an associated `Span` instance or via `Tracer`'s Extract capability.
+
+- **Set a Baggage item**, represented as a simple string:string pair. Note that newly-set Baggage items are only guaranteed to propagate to *future* children of the associated `Span`. See the diagram below. **(py: `set_baggage_item`, go: `SetBaggageItem`)**
```
yurishkuro: we also need to define baggage propagation rules for "sequence predecessor" reference. I assume the successor span inherits the baggage.

propagation.md
```diff
@@ -1,151 +1,147 @@
...
 
 <div id="custom-carriers"></div>
 
-## Custom Inject/Join Carrier formats
-
-Any propagation subsystem (an RPC library, a message queue, etc) may choose to introduce their own custom Inject/Join Carrier format; by preferring their custom format **but falling back to a required OpenTracing format as needed** they allow OpenTracing implemantations to optimize for their custom format without *needing* OpenTracing implementations to support their format.
+## Custom Inject/Extract Carrier formats
 
+Any propagation subsystem (an RPC library, a message queue, etc) may choose to introduce their own custom Inject/Extract Carrier format; by preferring their custom format **but falling back to a required OpenTracing format as needed** they allow OpenTracing implemantations to optimize for their custom format without *needing* OpenTracing implementations to support their format.
```
tinkerware: implemantations → implementations

spec.md
```diff
@@ -8,66 +8,96 @@ title: Semantic Specification
...
-                                [Span G······]
-                                [Span H··]
-                                        [Span I·]
+         [Span E·······]        [Span F··] [Span G··] [Span H··]
 ~~~
 
 ## Traces
 
-A **Trace** represents the potentially distributed, potentially concurrent data/execution path in a (potentially distributed, potentially concurrent) system. A Trace can be thought of as a tree of Spans. (See the ASCII diagrams above)
+A **Trace** represents the potentially distributed, potentially concurrent data/execution path in a (potentially distributed, potentially concurrent) system. A Trace can be thought of as a DAG of Spans. (See the ASCII diagrams above)
```
tinkerware: Would it be clearer if you said directed acyclic graph instead of using the acronym? It seems nitpicking, but not everyone in this document's target audience may have a comfortable grasp of graph theory concepts.

propagation.md
bensigelman: this typo is in `master`, good catch!

spec.md
bensigelman: sure, done

spec.md
bensigelman: I just did `s/children/descendants/`... the diagram makes this clear enough, I think.

use-cases.md
bensigelman: done

propagation.md
```diff
@@ -1,151 +1,147 @@
...
-
-# (If `carrier` held trace data, `span` will now be ready to use.)
+carrier = inbound_request.headers
+span_context = tracer.extract(opentracing.Format.TEXT_MAP, carrier)
+// Continue the trace given span_context. E.g.,
+//
+//     span = tracer.start_span("...", parent=span_context)
+//
+
+# (If `carrier` held trace data, `span_context` will now be ready to use.)
```
dawallin: I think this should still be span and not spancontext

spec.md
dawallin: Can the sequence predecessor reference also fan out and have multiple children? I guess it could.

spec.md
```diff
@@ -8,121 +8,162 @@ title: Semantic Specification
...
+**Sequence predecessor references:** Some Spans cannot start until a "sequence predecessor" Span finishes. For instance, chained sequence predecessor references neatly describe pipeline stages separated by queues.
+
+A sequence predecessor Span is part of the same logical trace as the descendant Span that refers to it: i.e., the descendant Span is formally caused by the work of its sequence predecessor.
+
+A timing diagram for a SequencePredecessorRef that must finish before the descendant Span can start:
+
+~~~
+    [-SequencePredecessorRef Span-]  [-New Span-]
+~~~
+
```
dawallin: I get confused here again, maybe because I don't see the use case for the different parent reftypes information. What are the different parent reftypes going to be used for? Why is it important to know if the parent span has finished or not?

In the go-tracer pull request https://github.com/opentracing/opentracing-go/pull/82 we discussed "Fire-and-forget", "Blocking children" and "Pipeline". Logically I saw "Fire-and-forget" and "Pipeline" group together as some event driven flow and "Blocking children" as their command driven opposite. Now "Fire-and-forget" is grouped together with "blocking children" instead which makes little sense to me.   

I thought what we were trying to capture was some kind of command vs event reftypes. I imaging that a remote client can know if it is blocking its parent or if it is part of a pipeline but not if its parent finished its span or not. And as a parent I think it makes sense to be able to start the next step in the pipeline, without being forced to finish my span first.

propagation.md
bensigelman: you are correct! good catch (bad find+replace on my side)

spec.md
bensigelman: yes, it could fan out... in my experience that's been less common than having each step in the sequence be its own "mini-root-span" where it acts as a `Parent` to any children that execute in parallel. I thought it best not to introduce that level of complexity to the diagram, though.

spec.md
bensigelman: @dawallin To me, the quintessential "fire-and-forget" request is a cache insert... no need to block the application path, and no need to do anything (aside from log failure) if it doesn't work. A pipeline is different, of course... there are typically queues (kafka, SQS, or even in-memory queues, whatever) between stages, and the application logic breaks if stages don't happen or fail (unlike fire-and-forget cache inserts, etc).

Also, the term "blocking" was removed from the Go PR, mainly for these reasons.

At issue here is our desire to keep the number of reference types as small as we can (since it's easier to add them than to remove them) while still adding value via these reference types. In my experience, the sequential pipeline case is the most important one to distinguish... it has its own heuristics from a UI/presentation standpoint as well as from a trace assembly and critical path standpoint. So it's the one I think we can afford to split off. As we gain more experience with reference types, perhaps we can split off common patterns into their own smaller types/subtypes of the Parent and SequentialPredecessor reference types.

Thoughts?

dawallin:
@bensigelman 
My biggest concern is about the sequential predecessor reftype been thightly connected to the span lifetime and the requierment that the parent span MUST been finished by the parent. 

Lets take an order service as an example. It will publish an event when an order is placed. It has no information who is interested and what will happen. In a parent perspectiv it is "fire-and-forget". It has two possibilities, either publish the event first followed by finishing the span, or the opposite, finish the span first and publish the event after (which sound strange to me but could be possible). Imagine a payment service reacting on this event. It will act as a tracing client and has the option to set a parent reftype. In the clients perspective the parent is a sequential predecessor, it continiues where the parent left, but it has no knowledge about, and is not interested in, in which order the parent published the event and finished the span. The logic at the parent finished at the publishing of the event, that is the important part, not whether the span finished or not.

Maybe this scenario is different from the sequential predecessor you have in mind and that it can be formulated as its own later when we have more experience, but to me it logically belongs more to the sequential predecessor than to the standard parent reftype.

bensigelman:
@dawallin I see your point re the possibility for time overlap near the finish of the sequential predecessor and the start of the descendant span. You've convinced me that the language cannot be as stringent about the time ordering of those two events (the predecessor's finish and the descendant's start).

Regarding what we can do about it: I still think there are essentially two cases here. The distinction is whether the referenced Span logically depends on the result of the descendant Span. If it does, it's a "parent"; if it doesn't, it's a... it's a... "predecessor"? I don't like that naming choice since the "parent" is also a "predecessor" from a strict timing standpoint. "Sequence" doesn't make as much sense to me when the two span lifetimes aren't disjoint in time.

Dependency on Descendant | No dependency on Descendant | Notes
---|---|---
Parent | SequencePredecessor | status quo
Parent | Predecessor |
Parent | Previous |
DependentAncestor | IndependentAncestor | verbose
Dependent | Independent | "independent" isn't quite the right idea
Up | Left | a la X-Trace
CausedBy & DependsOn | CausedBy | make the refs a bitmask
Cause & Dependent | Cause | variant on ^^^
DependentCause | Cause | another variant on ^^^, no bitmask

Thoughts from others welcome at this point.

spec.md
dkuebric: @dawallin @bensigelman it sounds like the distinction you're talking about is blocking vs nonblocking, but that's not what the ref type is handling.  For blocking vs nonblocking we use a special tag called "Async" that basically says "this span is asynchronous with respect to its parent and should not necessarily be considered blocking in the request flow".  Noting blocking vs nonblocking is useful for aggregate analysis of latency contribution (even if it's hard to get 100% accurate as sometimes you might join on some async work later in the trace...).

On the other hand, @bensigelman I'm wondering if "sequence predecessor" is necessary?  What information are you trying to capture there, or rather how would tracers use it in analysis?  It sounds like maybe this is similar to what I've been calling "meta-traces"--there are multiple units of work being chained together, but each one is kind of a separate trace with respect to aggregate timings that you'd want a tracer to capture?  Or other?

propagation.md
```diff
@@ -1,73 +1,74 @@
...
 - That said, the propagation mechanism should be extensible for optimizations
 
-## The basic approach: Inject, Join, and Carriers
+## The basic approach: Inject, Extract, and Carriers
 
-Any Span in a trace may be **Injected** into what OpenTracing refers to as a Carrier. A **Carrier** is an interface or data structure useful for inter-process communication (IPC); that is, the Carrier is something that "carries" the tracing state from one process to another. The OpenTracing specification includes two [required Carrier formats](#required-carriers), though [custom Carrier formats](#custom-carriers) are possible as well.
+Any SpanContext in a trace may be **Injected** into what OpenTracing refers to as a Carrier. A **Carrier** is an interface or data structure that's *useful* for inter-process communication (IPC); that is, the Carrier is something that "carries" the tracing state from one process to another. The OpenTracing specification includes two [required Carrier formats](#required-carriers), though [custom Carrier formats](#custom-carriers) are possible as well.
```
dkuebric: Is a carrier "useful" for IPC, or is it "used" for IPC?  I think this could be clearer, maybe along the lines of "A carrier is something used for IPC which can be extended with a metadata payload (e.g. HTTP request)."

spec.md
bensigelman: @dkuebric I've seen a common pattern where users end up with traces like this:

```
-----> time

[ Span A ]     [ Span B ] [ Span C ]            [ Span D ]
```

Between each span is a distributed queue (kafka, SQS, whatever), and the user cares about the time between the start of Span A and the finish of Span D.

dkuebric: Yeah this is an interesting use-case.  A single ID across all 4 stages has the advantage of grepability in logs, which is nice.

However, in our work we are leaning towards modeling this slightly differently.  One thing we found in trying to support cases like this is on the tracer collector/analysis side, it's much easier if the "single trace" that spans all 4 of those spans is actually 4 separate traces which refer to each other via a "parent trace" reference.  This helps us figure quickly out when "done" happens for each segment--a need that might be specific to the types of analysis we do.

The two approaches require about the same level of consideration on the instrumenter's part: the insight that the unit of work in each subsequent parent span is decoupled, and therefore the trace in the process should be started not as a standard continuation of an existing trace but as either a "sequence predecessor" reference or "parent trace" reference.

So these are very nearly 6 and 1/2 dozen (just different words) but the difference is whether the trace ID changes (or can change) at the juncture as well.  Curious what you think of the multi-ID model?

dkuebric:
@bensigelman the answer to the terminology question for the approach I proposed is that if they are totally separate, there's separate traces, and so the data type changes from a parent span to a parent trace depending on the case.

yurishkuro:
I think we're talking about two dimensions here (at least). In a classic DAG every node maps just one point on the timeline, so for example X-Trace has no issue representing causality between any two nodes. In case of a DAG constructed of spans, every node has two points on the timeline, so the causality relationship (edge) can be labeled in several ways (cf. [Microsoft Project "link types"](https://support.office.com/en-us/article/Create-task-dependencies-links-within-your-project-71fa4768-bf0d-414a-96e7-4e5d90df9676) - fwiw, I never thought they are intuitive enough).

Then there is a separate "family" dimension as "parent-child" vs. "siblings/sequence" relationships. They also existed in X-Trace, as "down" vs. "next". It is possible to think about these without considering timeline causality (except that some timeline relations would not make sense, such as child starting before parent).

So maybe we need to model these two dimensions separately, as two fields in the span reference.

Again, certain permutations of two fields may not make sense, but at least we'll be covering all available space, rather than trying to enumerate only those combinations that we thing do make sense.

spec.md
bensigelman: Of course the OT layer per se does not know or care about IDs... since you're asking, though, I would only _want_ one ID for the trace of a sequence of operations that have explicit causal relationships. I like multi-ID tracing when distinct traces end up with a shared fate due to a shared resource (a queue, batching, whatever).

propagation.md
bensigelman: I meant "useful" in that the literal data structure used for IPC is rarely going to be general enough... since we came up with that terminology, though, the "carrier" has become less of an object/struct and more of an abstract interface for reading/writing a certain type of data (e.g., raw bytes, key:val pairs, or http headers).

I want to sleep on this, but I can imagine `s/carrier/projector/` or something (the interfaces have started to look more like that lately anyway).

bensigelman:
@yurishkuro I agree that timing is an important aspect, but we already capture that with the start and finish timestamps... so it's really just the parent/child vs sibling/sequence aspect I'd like to capture here. Ideally without boiling the ocean or adding more than a very short enum since this is going to be very hard to get right the first time.

yurishkuro:
@bensigelman start/finish timestamps do not capture causality. E.g. X started after Y - was it just a race condition or a true dependency.

I think one of the points @dawallin had was that we're mixing "family" relationships with blocked-by/started-after, which are timeline causality dimension.

tinkerware:
Would using a separate axis of "family" relationships really make sense when thinking about spans? It seems that, rather than being an independent set of concepts, they are synonyms for different causality relationships. For instance, parent-child relationship is happens-before plus depends-on (let's assume that happens-before has a clear and useful definition here). Sibling relationship, likewise, is a synonym for concurrent spans.

bensigelman:
@yurishkuro @dawallin @dkuebric @tinkerware 

I'm trying to think about the axes here in a way that's forwards-compatible AND simple enough to make an OpenTracing newbie's head spin. Hopefully there's still something left in that Venn diagram :)

Let's call the span we're referencing the "ancestor" (in Dapper/Zipkin, the "parent"), and the span we're starting the "descendant". For this rev, I do not care about cases where the descendant isn't *caused* by the ancestor... I consider those cases beyond the core causality-tracking bread and butter of distributed tracing. That said, I'm convinced that it's important to distinguish whether the ancestor depends on the outcome of the descendant (similar to the "async" case @dkuebric described). I don't believe there is anything else worth distinguishing at the moment.

So my concrete proposal:
- make it explicit that OpenTracing "references" _only_ capture causal relationships in this rev
- only allow the user to distinguish between "Ancestor depends on Descendant" and "Ancestor doesn't depend on Descendant". If I had to come up with names (from the perspective of the descendant), I'd call it `Parent` and `Predecessor`... the comments can be wordier, but I'm concerned about overwhelming developers who don't have time to RTFM if we choose more jargony terms.

Thoughts? Esp specific counter-proposals since there are a lot of tradeoffs here.

PS: @yurishkuro, I misunderstood what you were saying earlier... I certainly agree that time ordering on its own doesn't guarantee any particular causal relationship.

tinkerware:
How about `precedes` to indicate the ancestor does not depend on descendant, and `subsumes` to indicate that the ancestor both causes and depends on the descendant?

bensigelman:
@tinkerware just to make this more concrete, the code would might look roughly like this:

```py
# longhand version:
descendant = tracer.start_span("operation", references=[opentracing.parent(span_context)])

# shorthand version:
descendant = tracer.start_span("operation", parent=span_context)
```

or

```go
descendant := tracer.StartSpan("operation", opentracing.Parent(spanContext))
```

The way you've written things, it would be from the perspective of the ancestor (which "subsumes" the descendant): i.e., the converse of what's needed at the StartSpan call site.

tinkerware:
@bensigelman ah, this grounds things for me better, thanks.

How about this:

```go
# when ancestor does not depend on descendant, as in async use cases
descendant := tracer.StartSpan("operation", opentracing.Follows(ancestorContext))
```

or 

```go
# when ancestor depends on descendant for completion
descendant := tracer.StartSpan("operation", opentracing.Joins(ancestorContext))
```

I think these two terms get across the difference between the two roles clearly. I chose to name the _edge_ rather than the _vertex_ here; that meaning seems to be closer to the perspective of a reader trying to grasp the code at the call site and matches better with the syntax that we are assuming will be commonly used.

As a side benefit (for a use case you've ruled out in the initial scope), you can better describe the relationship between ancestor spans for individual operations and a descendant that represents a coalesced operation, like a group commit:

```go
# when descendant coalesces multiple ancestors
descendant := tracer.StartSpan("operation", opentracing.Groups(ancestors))
```

There are subtle semantics here (do the ancestors, in turn, depend on the grouped descendant? if not, what, if anything, can OT do to link the trace of a background task with potential ancestors) and I agree that it is early to try to resolve these questions now.

bensigelman:
@tinkerware I like `Follows`... clear, plain-english, and descriptive. `Joins` seems too general (or too easily misunderstood) for what we're getting at here, though. I wonder if `For()` would work? Or `OnBehalfOf()`?

tinkerware:
We could try and be more explicit, for example `JoinsBackTo`. But I understand where you are coming from; it deemphasizes the sense of being caused by, and feels more general. How about `BelongsTo`?

bensigelman:
I like `BelongsTo`, personally... also clear, also plain-english, also descriptive.

spec.md
```diff
@@ -8,121 +8,162 @@ title: Semantic Specification
...
-
-Note that `Inject` and `Join` are not entirely symmetrical interfaces since `Span` injection is a fundamentally lossy process (most of the `Span`'s state is meant to be recorded out-of-band from the application's own communication path, whereas Inject-ed state is meant to be sent in-band along with the application data).
-
+- **Start a new `Span`**. The caller can specify zero or more [`SpanContext` references](#references) (e.g., parents or sequence predecessors), an explicit start timestamp (other than "now"), and an initial set of `Span` tags. **(py: `start_span`, go: `StartSpan`)**
```
tinkerware: We can change here to say:

> The caller can specify zero or more [`SpanContext` references](#references) (e.g., with `BelongsTo` or `Follows` roles).

We'd have to change the `#references` section to explain these. Nice thing about `BelongsTo` is that we could still talk about the ancestor as a _parent_ owning its child spans, which in turn belong to their parent.

bensigelman: yeah, there are a bunch of misc places that will have to change in the actual PR... just wanted to come to agreement about terminology before I spent time tweaking the .md files.

yurishkuro:
This thread needs a principle component analysis :-)

We are still taking about the timeline relationship of two segments, A(start/end) -> B(start/end). Because we state they are ancestor/descendent, we have an invariant `A.start < B.start`. The other permutations are:
  * parent: `B.start < A.end`
    * special case - blocking parent: `B.start < B.end < A.end`
  * pipeline: `A.end < B.start < B.end`
  * join: similar to pipeline but for A1...An

Are there any other aspects we want to capture besides the timeline relationship?

bensigelman:
@yurishkuro my latest thinking doesn't define this stuff in terms of timelines as much as information dependencies: does `A` depend on the the work done by `B`? If so, `B` belongs to `A` (its parent). Otherwise, all we can say is that `B` follows `A`, and I'm no longer asserting that `A.end < B.start` thanks to some of the comments above that set me straight on that front.

BTW, I don't want to say that `A` "blocks" on `B`, just that it depends on `B` in some way... e.g., an RPC client span (`A`) may have a client-side timeout and thus end before the server span (`B`) from a sequencing standpoint.

PS: I may refrain from further PR commenting on this since we have our weekly hangout at 1:30pm PT / 4:30pm ET and it will be easier to reach consensus there.

@dawallin @dkuebric @tinkerware https://plus.google.com/hangouts/_/calendar/YmhzQGdtYWlsLmNvbQ.hb23eot6tj66quam081ch51ga4?authuser=0 is the call link if you're available then.

tinkerware:
I'm hesitant to see this spec impose constraints on timeline relationships (except the most basic ones, such as for any event end timestamp can not be before the start timestamp), which I see as different from causal relations. They are bound to be guidelines rather than actual constraints in the context of distributed tracing, and will require implementations to handle situations that fall outside those guidelines. Implementations may also end up in situations where they have no choice but to generate traces that violate those guidelines (e.g. referencing spans extracted from a node with clock skew). 

Is there any need for tracers and instrumenters that would not be met by stating causality constraints rather than timeline?

yurishkuro:
> does A depend on the the work done by B? If so, B belongs to A (its parent). Otherwise, all we can say is that B follows A.

By that definition an async cache insert is no longer a child of A. Put differently, above sounds like a "blocking parent".

Note that I am not suggesting that we use timeline dependencies as THE definition. But they do imply information dependencies. As we're having difficulties clearly separating those information dependencies,  specifying them in terms of timeline dependencies make it less ambiguous. Without timeline, have `parent` (special case `blocking parent`) and `join` (special case `pipeline`). 

I don't mind how we call them, I want the space partitioning to be non-overlapping and to make sense. And be easily explainable to newcomers (I think timeline explanations are very obvious).

spec.md
adriancole: where can I find out more about "sequence predecessors"? I don't see it in the use cases doc although I can grok it. Particularly interested in how folks use the OT to annotate the sequence relationship. I'm also interested in how folks represent this in implementation. For example, I know Naver use a "nextId" concept to help, although depending on the semantics of sequence predecessor, it may not be the same thing.

https://github.com/naver/pinpoint/blob/d42fb8437a4098ed375790810c5a30452f05074d/bootstrap-core/src/main/java/com/navercorp/pinpoint/bootstrap/context/AsyncTraceId.java#L11

cc @xylus

spec.md
adriancole: passing note.. there's a fairly important new concept kited in this PR, which one can familiarize with by reading the go implementation. Maybe the semantics of this can be in a separate topic or PR in this repo? Looking at the title, this PR is basically a rename.. looking at the comments, you see a powerful new tool discussed (sequential predecessor etc)

adriancole: wrt "it's much easier if the "single trace" that spans all 4 of those spans is actually 4 separate traces which refer to each other via a "parent trace" reference." This was also called as "soft links" at the last workshop by @bogdandrutu. 

https://docs.google.com/document/d/1iWy3LrMLj329dCAiTg_mM9eP2s1B7KM4stg_8EfAmFM/edit#heading=h.auth91bysajd

This stuff is so conceptually important, I think it warrants being discussed (maybe allowing an hour or more) at the next workshop, too, even if the decision here is already made by then.

spec.md
adriancole: ps here's a note I found on sequence https://github.com/openzipkin/zipkin/issues/780#issuecomment-149117067

spec.md
dkuebric: @adriancole I missed that "soft link" terminology, thanks for sharing.  At AppNeta we've been calling them parent/child traces.  I do think it's a great simplifying assumption for trace implementations (vs having to consider traces that span arbitrarily long spans of time and computation), agree seems like an interesting topic for the workshop.

bensigelman: @adriancole I agree 100% that the "reference" concept is important, and also acknowledge that it's a bit buried in this PR... my goal here is to come up with the absolutely minimal set of reference types to start with and wait until we have more experiences to fill out the enum/matrix. That "wait"ing will certainly extend past July 28, and I think this would make an excellent topic for the workshop, too :)

spec.md
bensigelman: @adriancole @xylus the concept "sequence predecessors" ended up evolving (in this PR discussion and in our OT hangout today with @yurishkuro and @dkuebric) into something more like `Follows()` per @tinkerware's comment here: https://github.com/opentracing/opentracing.github.io/pull/100#issuecomment-229404325

That whole thread is pretty interesting IMO.

adriancole:
here's what I can grok so far..

sequence is important, particularly as it allows you to visualize intended order. This comes up often enough, ex in pipeline or workflows, and also when people are debugging scheduling delay etc. This has been very discussed even before OT.

Dependence seems also useful, as it could help triage problems, but it isn't limited to things on a straight line. For example, regardless of if you are creating a child or a next sibling span, in either case, you could choose to not care about the result (async). For example, the last stage in a pipeline could be fire and forget.

If above makes sense, then if is one easy to define than the other. Why don't we fork an issue for the one easier to pin down and make it separate from the other matters.. It would be helpful as it takes hours to grok all these things, and more concepts makes issues like these less approachable and/or more expensive.

Specifically, let's fork off how to indicate next child vs next sibling in some way that doesn't require reading the golang repo?

bensigelman:
I pushed some updates to the actual PR that started this (great) discussion... PTAL? I think this is consistent with the discussion at the weekly hangout today. Not 100% sure I like the names, though I am satisfied with the concepts for this rev.

spec.md
```diff
@@ -8,121 +8,160 @@ title: Semantic Specification
...
+
+<span id="references"></span>
+
+### Causal Span References
+
+A Span may reference zero or more Spans that are causally related. OpenTracing presently defines two types of references: `BelongsTo` and `Follows`. **Both reference types specifically model direct causal relationships between a descendant span and an ancestor span.** In the future, OpenTracing may also support reference types for spans with non-causal relationships (e.g., Spans that are batched together, Spans that are stuck in the same queue, etc).
+
+**`BelongsTo` references:** A descendant Span may "BelongTo" an ancestor Span. In a "BelongsTo" reference, the ancestor Span depends on the descendant Span in some fashion (often—but not always—the ancestor Span cannot finish until the descendant Span finishes). We sometimes refer to the "BelongsTo" ancestor as the "parent" and the descendant as the "child".
```
adriancole: DependsOn is used to define what BelongsTo means. Maybe use DependsOn?

spec.md
```diff
@@ -8,121 +8,160 @@ title: Semantic Specification
...
+
+A typical timing diagram for a descendant Span that BelongsTo a ancestor Span:
+
+~~~
+    [-Ancestor (Parent) Span-----------]
+         [-Descendant (Child) Span----]
+~~~
+
+**`Follows` references:** Some ancestor Spans do not depend in any way on the result of their descendant Spans. In these cases, we say merely that the descendant Span "Follows" the ancestor Span in a causal sense. There are many distinct "Follows" reference sub-categories, and in future versions of OpenTracing they may be distinguished more formally.
```
adriancole: Not terribly tied to this, but HappenedAfter ties nicely with HappenedBefore which seems semantically similar relationship?

spec.md
adriancole: oops it is the inverse.. dependentOf, or more simply childOf

use-cases.md
```diff
@@ -153,8 +153,9 @@ func BusinessFunction1(ctx context.Context, arg1...) {
 }
 
 func BusinessFunction2(ctx context.Context, arg1...) {
-    parentSpan := SpanFromContext(ctx)
-    childSpan := opentracing.StartChildSpan(parentSpan, ...)
+    parentSpan := opentracing.SpanFromContext(ctx)
+    childSpan := opentracing.StartSpan(
+        "...", opentracing.BelongsTo(parentSpan.Context()), ...)
```
adriancole: `belongs_to` is interesting as it doesn't imply a hierarchy, and maybe I could put my grandparent there, because I indirectly do.. Not sure if this flexibility is what was intended or not.

use-cases.md
```diff
@@ -206,16 +207,16 @@ def traced_request(request, operation, http_client):
...
 
-### Using Distributed Context Propagation
+### Using Baggage / Distributed Context Propagation
 
 The client and server examples above propagated the Span/Trace over the wire, including any Baggage. The client may use the Baggage to pass additional data to the server and any other downstream server it might call.
 
 {% highlight python %}
 # client side
-span.set_baggage_item('auth-token', '.....')
+span.context.set_baggage_item('auth-token', '.....')
```
adriancole: span exposes a reference to its context now? or was that the case before this PR

dawallin: @adriancole That is where all this started. PR "introduce and take advantage of SpanContext"  https://github.com/opentracing/opentracing-go/pull/82. This PR is the spec update for that one.

dawallin:
LGTM. This resolves the concern I had with the ref types in the beginning. Good discussion.

use-cases.md
adriancole: thx for the tip!

tinkerware:
LGTM 🚀

spec.md
```diff
@@ -8,121 +8,160 @@ title: Semantic Specification
...
+- <span id="inject-extract"></span>**Inject a `SpanContext`** into a "carrier" object for cross-process propagation. The type of the carrier is either determined through reflection or an explicit [format identifier](/propagation#format-identifiers). See the [end-to-end propagation example](/propagation#propagation-example) to make this more concrete.
+- **Extract a `SpanContext`** given a "carrier" object whose contents crossed a process boundary. Extract examines the carrier and tries to reconstruct the previously-Injected `SpanContext` instance. Unless there's an error, Extract returns a `SpanContext` which can be used in the host process like any other, most likely to start a new (descendant) Span. (Note that some OpenTracing implementations consider the `Span`s on either side of an RPC to have the same identity, and others consider the caller to be the ancestor and the receiver to be the descendant) The type of the carrier is either determined through language reflection or an explicit [format](/propagation#format-identifiers). See the [end-to-end propagation example](/propagation#propagation-example) to make this more concrete.
```
dominikh: > Note that some OpenTracing implementations consider the `Span`s on either side of an RPC to have the same identity, and others consider the caller to be the ancestor and the receiver to be the descendant

For implementations that consider the receiver to be the descendant, is Extract supposed to return a new SpanContext with a new span ID, descending from the extracted parent context?

bensigelman: no, `Extract` should really just extract whatever context was injected and sent over the wire... the user would then `StartSpan` and pass that extracted context in as a `Parent` or `BelongsTo` or `DependentOf` or whatever we end up calling it :)

use-cases.md
bensigelman: @adriancole we should talk about it IRL in July (or before over VC, whatever)... an interesting evolution to this point, but I think exposing an explicit context object solves lots of corner-case problems and points of confusion with OpenTracing, esp around propagation, more exotic span-to-span references, and use-after-finish timing issues with Spans.

adriancole: sounds good. let's rope off time late july!

spec.md
bensigelman: I am worried it's just because of the multi-year brainwashing of dapper/zipkin/etc, but it is awfully hard to shake the parent/child naming concept... so `ChildOf` sounds natural to me. Any other thoughts about `ChildOf` vs `BelongsTo`? I also considered `NeededBy`, but that implies that the ancestor literally blocks, which is not necessarily the case per the explanation above.

adriancole: When I think of ChildOf vs Follows (or HappensAfter), brain washing makes me think of implementing the first as a child, where the latter could just be a sibling.

When I think of BelongsTo vs Follows (or HappensAfter), brain washing makes me wonder if BelongsTo is referring to the trace (understand trace id isn't necessarily defined here) or the span component? I don't necessarily get to a child relationship without some nervousness or reading more carefully.

ack that dapper/zipkin brainwashing shouldn't anchor things too much, but wondering if the prevalence of vocab like "child spans" in many systems might warrant just going more status quo

yurishkuro: my meter is certainly in the red on `BelongsTo`. I'm ok with `ChildOf` if we don't have anything better - it is at least intuitive, even if inaccurate since what we're capturing is parent's dependency on the child's outcome. `DependencyOf` would be more accurate term (but long and ugly to type).

FWIW, parent/predecessor (especially as _options_ attached to span) feel cleaner than childOf/follows.

tinkerware: I'm ok with `childOf`/`follows` instead of `belongsTo`/`follows`. `BelongsTo` meshes better with `follows`, but @adriancole's point about the prevalence of parent/child vocabulary is worth consideration.

I find `parent`/`predecessor` phrasing clunky compared to the alternatives we have. It also seems harder to extend to new relationship roles.

dkuebric: I think child/parent is accurate for describing both cases, because what we're describing is a tree of spans.  The graph/tree relationships concept is something a lot of users would probably be familiar with.

So in the vein of trying to preserve/harness that existing semantic mapping, what if both proposed types are child relationships, but there's two types: one is `dependent` and one is `independent`?

bensigelman: The good and bad news here is that all of these various viewpoints are valid and defensible, and none is a perfect fit.

For the "dependent" case, I am going to go with `ChildOf`. Rationale:
- I think it's important for the term to feel "transitive" in the sense of a "transitive verb": these are directed edges we are decorating, and it must be clear which side is which. @yurishkuro this argues against `Parent(...)` where I think it's potentially confusing whether the span one is starting is a parent or a child. `ChildOf` doesn't have that problem.
- Whether due to brainwashing or not, it is conventional in the distributed tracing universe and being unsurprising is a big plus.
- After sleeping on it, I'd argue that `BelongsTo` implies containment more than causal dependency... also feels jargony per the next bullet
- There are more precise terms (`DependencyOf`, for instance) but I think they are too jargony and probably too verbose as well


For the "independent" case:
- @dkuebric makes the point that both reference types could be considered "children"... I see and recognize that point, but from a drill-down / critical path standpoint, I'd argue that the "dependent" case is more parent-child like than the independent case which is more like a sibling relationship, esp for pipelines.
  - Also, as a parent of an 19-month-old, "Independent Child" is just too much of an oxymoron for my current life stage :)
- `Predecessor` has the same problem as `Parent` from a directed-edge standpoint
- `SuccessorTo` feels to jargony
- `HappensAfter` is ok with me... I wish that it explicitly ruled out the dependent cases, though, as it sounds like `HappensAfter` is a superset of `ChildOf` when it is not intended to be. It may also imply that the new span starts after the ancestor *finishes*, which is not necessarily the case, of course.
- `Follows` is also ok with me, thought it would be nice to match the `<Noun><Preposition>` form of `ChildOf` for consistency. It also has the problem about not ruling out parent-to-child dependencies.
- `SpawnedBy` or `DueTo` might do a better job expressing the independent nature of the newly started span... though they fall into the jargony bucket.

I'll try to make an actual decision and send out an update later today.

dkuebric: @bensigelman I'll make a second case for the "it's always parent/child": easy defaults.  Specifying the type of a relationship could be an optional step, making the instrumentation learning curve something like this:

1. Learn how to associate spans as parent/child.  Everything works, though if the tracer has fancy critical path analysis features those might be slightly off.
2. If you are doing automated critical path analysis and have fire-and-forget or other async stuff, annotate some of them as "independent" or whatever terminology.  Now the extended features work.

I realize I'm making a proposal about what's a "basic" vs "advanced" feature in a tracer, but I think this is a place where we can both provide for formal correctness but also let that take an optional role in the interest of having a shallow learning curve.

yurishkuro: In the "causal DAGs" literature, the parent/child relationship is well defined: if node A causes node B, then A is a parent and B is a child. This agrees with what @dkuebric is saying, that _all_ relationships we discussed here are parent/child, including the pipeline case `[A]->[B]->[C]`. I think that also explains why the terminology is prevalent in distributed tracing. I don't think this general terminology necessarily implies "parent depends on child". So I rather like Dan's suggestion of `ChildOf` being the most general span reference type, which will cover the majority of use cases. We tried to have clear space partitioning with the enums, but `ChildOf` is already a partitioning, the opposite side being "no parent". So the other enums could be sub-partitions of the `ChildOf` space. E.g. `AsyncChildOf`.

bensigelman: Everyone is so persuasive... :)

I was trying to rewrite the document to use some term other than "ancestor" and "descendant" but failed... basically underlining the point that everything is parent-child, formally.

I remain attached to `ChildOf` for the parent-depends-on-child case.

For the other case, I don't like the term `Async` as "async" shouldn't imply that there's no dependency: plenty of parent spans keep on doing work (hence have spawned a child "async") but cannot complete until the child is finished and are thus "dependent". I could go with `DetachedChildOf` or `OrphanChildOf` or `OrphanOf`.

yurishkuro: `DetachedFrom` ?

tinkerware: `SplitFrom`? `Forked` (ok, maybe not that one)

tinkerware: If we are saying that it's turtles all the way down, then `ChildOf` is information-free; we can instead try and name the additional attributes of the relationship that provides the distinction. Practically speaking, the API for creating a child span does not have to require any role/relationship name in the happy case, which is caused-by/depends-on (or split/join, the way I prefer to think about it). 

Also, I notice that terms that focus on the _structure_ of the trace tend to miss the aspect of _flow_ that seems to underpin my thinking; the dynamic aspect of the execution we are tracing. Now that I think about it, that's probably the reason I liked `Follows` best.

Still thinking about something that would capture "split w/o join" or "no dependency". I agree that `Async` is not it.

yurishkuro: > If we are saying that it's turtles all the way down, then ChildOf is information-free;

It's not if you consider the case of starting the root span with *no* parent reference.

bensigelman: I just pushed a version of this that has `ChildOf` and `FollowsFrom`... I like those and would be fine merging this as such, though veto votes are welcome and I certainly recognize that this is really subtle. (As the length of the thread indicates)

bensigelman:
TL;DR the latest commit moves to `ChildOf` and `FollowsFrom`...

data-semantics.md
```diff
@@ -27,7 +27,7 @@ Secondarily, an optional **component** tag can be provided to scope the operatio
 
 ### Span Structure
 
-Span structure is also important: what do spans represent, and what relationship do they have towards their parent span?  This is covered in the [Semantic Specification](/spec).
+Span structure is also important: what do spans represent, and what relationship do they have towards their ancestor spans?  This is covered in the [Semantic Specification](/spec).
```
adriancole: supernit: can replace "ancestor spans" w/ "ancestors" since the word span is already mentioned

adriancole:
"FollowsFrom" works for me.. It doesn't confuse me, and I'm easily confused.

:shipit:

data-semantics.md
bensigelman: sure, done :)

dawallin:
I miss some words often associated with async programming. Await, Promise and Future. What about PromiseFor or AwaitedBy or FutureOf? They indicate that the parent expects something but is not necessary blocked.

dominikh:
Those async terms sound very specific to certain programming languages. E.g. in the Go community, a lot of people will not be comfortable with those terms.

bensigelman:
@dawallin per

> What about PromiseFor or AwaitedBy or FutureOf?

These async constructs all imply that the parent cannot continue until the child finishes, which is not strictly the case here. For example, an RPC client span (the parent) may have a 5-second timeout and thus terminate before the RPC server span (the child) actually finishes and returns its result.

Anyway, unless anyone specifically objects, I would like to merge what's presently in this PR... I'll do so on Monday, so speak now or forever hold your peace :)

In the meantime I will update the (intentionally still-pending) `opentracing-go` change to follow suit with the diff here.

propagation.md
```diff
@@ -1,73 +1,74 @@
...
-# The operation name we want for the server-side span:
-operationName = ...
-serverSpan = tracer.join(
-    operationName, opentracing.Format.TEXT_MAP, carrier)
+carrier = inbound_request.headers
+span_context = tracer.extract(opentracing.Format.TEXT_MAP, carrier)
+// Continue the trace given span_context. E.g.,
+//
+//     span = tracer.start_span("...", child_of=span_context)
+//
```
yurishkuro: # Python comments start with `#`
# line 70 does not need to be commented out

bensigelman: (good catch)
ᐧ

spec.md
```diff
@@ -8,121 +8,160 @@ title: Semantic Specification
...
+A **Span** represents a logical unit of work in the system that has a start time and a duration. Spans may be nested and ordered to model causal relationships. Each span has an **operation name**, a presumably human-readable string which concisely represents the work done by the span (e.g., an RPC method name, a function name, or the name of a subtask or stage within a larger computation).
+
+<span id="references"></span>
+
+### Causal Span References
+
+A Span may reference zero or more Spans that are causally related. OpenTracing presently defines two types of references: `ChildOf` and `FollowsFrom`. **Both reference types specifically model direct causal relationships between a child span and a parent span.** In the future, OpenTracing may also support reference types for spans with non-causal relationships (e.g., Spans that are batched together, Spans that are stuck in the same queue, etc).
+
+**`ChildOf` references:** A Span may be the "ChildOf" a parent Span. In a "ChildOf" reference, the parent Span depends on the child Span in some fashion (often—but not always—the parent Span cannot finish until the child Span finishes).
+
```
yurishkuro: suggest replacing _"(often—but not always—the parent Span cannot finish until the child Span finishes)"_ with:

Some common examples include:
 * the parent span is blocked until the child span returns (typical synchronous RPC case)
 * the parent span is doing work, but cannot finish until the child span returns (typical asynchronous RPC case)

spec.md
```diff
@@ -8,121 +8,160 @@ title: Semantic Specification
...
+A typical timing diagram for a Span that is the ChildOf a parent Span:
+
+~~~
+    [-Parent Span---------]
+         [-Child Span----]
+~~~
```
yurishkuro: Suggestion: per above, could be worth showing 2 diagrams, a parent with two sequential, blocking children, and a parent with two concurrent children.

spec.md
```diff
@@ -8,121 +8,160 @@ title: Semantic Specification
...
+As is the case with Logs, if certain known tag key:values are used for common application scenarios, tracers can choose to pay special attention to them. For more information, see the [Data Semantics Guidelines](/data-semantics).
 
-## Inject and Join
+## SpanContext
 
-Spans may be **Injected** into and **Joined** from **Carrier** objects that are used for inter-process communication (e.g., HTTP headers). In this way, Spans may propagate across process boundaries along with sufficient information to join up with the Trace in some remote process.
+Every Span must provide access to a **SpanContext**. The SpanContext represents Span state that must propagate to child Spans and across process boundaries (e.g., a `<trace_id, span_id, sampled>` tuple) and also encapsulates any **Baggage** (see below). SpanContext is used when propagating traces across process boundaries and when creating edges in the trace graph (e.g., ChildOf relationships or other [references](#references)).
```
yurishkuro: first line - "must propagate to _children_ Spans"

spec.md
```diff
@@ -8,121 +8,160 @@ title: Semantic Specification
...
 
 ### Global and No-op Tracers
 
-OpenTracing libraries must provide a no-op Tracer as part of their interface. The no-op Tracer implementation must not crash and should have no side-effects, including baggage propagation. The Tracer implementation must provide a no-op Span implementation as well; in this way, the instrumentation code that relies on Span instances returned by the Tracer does not need to change to accommodate the possibility of no-op implementations. The no-op Tracer's `Inject` method should always succeed, and `Join` should always return a "trace not found" status.
+OpenTracing libraries must provide a no-op Tracer as part of their interface. The no-op Tracer implementation must not crash and should have no side-effects, including baggage propagation. The Tracer implementation must provide a no-op Span implementation as well; in this way, the instrumentation code that relies on Span instances returned by the Tracer does not need to change to accommodate the possibility of no-op implementations. The no-op Tracer's Inject method should always succeed, and Extract should always behave as if no `SpanContext` could be found in the carrier.
```
yurishkuro: > The Tracer implementation must provide a no-op Span implementation as well; in this way, the instrumentation code that relies on Span instances returned by the Tracer does not need to change to accommodate the possibility of no-op implementations.

Not sure what this is about (although it's been here before). I don't know of scenarios where a real tracer would need to return a noop span.

use-cases.md
```diff
@@ -86,44 +86,44 @@ def top_level_function():
...
-    operation_name=operation,
+extracted_context = tracer.extract(
     format=opentracing.HTTP_HEADER_FORMAT,
     carrier=request.headers
 )
-if span is None:
+if extracted_context is None:
     span = tracer.start_span(operation_name=operation)
+else:
+    span = tracer.start_span(operation_name=operation, child_of=extracted_context)
```
yurishkuro: Q: could we allow `child_of=None` (Go: `childOf(nil)`)? Before this change it was certainly convenient to be able to do above in just one line.

yurishkuro: Also, I think we need to change this to use the `RPCParent` reference, so support Zipkin's one span_id per RPC model. Currently this example breaks Zipkin tracers (and it's not using std tag constants).

yurishkuro:
LGTM

spec.md
bensigelman: If there was `github.com/english/issues` it would be overflowing with complaints :)

I think this is correct, but it's admittedly a corner case. We are using the word "child" as an adjective here, not as a noun (as it would typically be used, that is), and that changes the rules... Imagine that we only talk about propagating to, say, Spans based on their color; then of course we would write

> "... Span state that must propagate to *red* Spans and across process..."

In this sentence, since "child" is a modifier on Spans, I think (??) what I have is correct. Though note that

> "... Span state that must propagate to child and across process ..."

is (clearly) wrong. *That* would indeed be

> "... Span state that must propagate to children and across process ..."

Anyway, YMMV and I am only like 80% confident about this.

spec.md
bensigelman: Ah, "OpenTracing libraries" here are `opentracing-go`, `opentracing-java`, etc... the point is that there needs to be *some* builtin `Tracer` impl that can return something "safe" from `StartSpan`.

I'll clarify.

use-cases.md
bensigelman: > Q: could we allow child_of=None (Go: childOf(nil))? Before this change it was certainly convenient to be able to do above in just one line.

I can see it either way... I'd like to not-specify that as part of this change, though (more of a per-platform thing).

bensigelman: re `RPCParent`: I see what you're getting at, but I'd like to revisit this once we've hashed through the implications of all of this in `opentracing-python` if that's ok with you. Otherwise we'll just end up changing this example twice.

bensigelman:
@yurishkuro thanks for taking such a close look, I appreciate it. I have pushed some updates.

PR #118, opened 2016-09-09
How-to doc for instrumenting existing systems

azimman:
In conversations with OT users, various people have said that the function-level documentation and semantics are clear, but the big picture of how to approach instrumentation for large codebases is less so. 

The goal of this document is to help individuals and teams that are looking to instrument existing environments with OpenTracing. It is written as a 'how-to' format with a simple example in an effort to be more approachable to developers lacking prior experience in distributed tracing.

Feedback is welcome.


systems-at-scale.md
```diff
@@ -0,0 +1,73 @@
...
+_Before getting onto recommendations on how to instrument your large-scale system with OpenTracing, be sure to read the [Specification overview](/spec)._
+
+<div id="toc"></div>
+
+
+## Spans and Relationships
+
+The two fundamental aspects of implementing OpenTracing across your infrastructure are _Spans_ and the _Relationships_ between those spans:
+
+* **_Spans_** are a logical unit of work in the system that has a start time and a duration. In a trace Spans are the components across the systems exercised on a specified path.
```
yurishkuro: maybe "Span is a logical unit of work"

yurishkuro: the second sentence is odd, spans are not components.

azimman: How about:

* **_Span_** is a logical unit of work in the system that has a start time and a duration. In a trace Spans are associated with the components of the system as they are exercised on a specified path.

yurishkuro: On second thought, why do we need to re-define span & reference here again?  I think it only leads to more inconsistencies in the docs, I would prefer all definitions to be in one place, and currently it's the spec.md

rpc-ipc.md
```diff
@@ -0,0 +1,190 @@
...
+    * Extract the current trace state from the inter-process transport (HTTP, etc)
+    * Start the span
+    * Store the current trace state
+* Server Finishes Processing the Request / Sends Response
+    * Finish the span
+
+Because this workflow depends onrequest processing, you'll need to know how to  change the  framework’s requests and responses handling--whether this is through filters, middleware, a configurable stack, or some other mechanism.
+
+### Extract the Current Trace State
+
```
cwe1ss: Missing space in "onrequest"

azimman: fixed.

systems-at-scale.md
azimman: @yurishkuro you make a good point around consistency. This is something that I have been trying to pay close attention to. In the case of this document the terms are not being re-defined, instead the intent is to use the definitions in context. 

I am actually planning to tackle the spec.md page next and consistencey will be a major focus. This will be part of the effort to make the overall content more organized and approachable. 

I'm not sure that removing the definitions and diagrams from the opening of this document makes it more digestible. In my opinion, the reiteration helps to frame and focus the document. would you be open to leaving it in for v1 and if you still feel that it is redundant after the updates to spec.md we can take it out?

pritianka: Hi Guys! One thing to note - I noticed gRPC and some other open source projects always define the project in the beginning of documents (regardless of repetition). In our case, spans and their relationships is a key concept for the best practices doc, so it will be a good memory jog for people to read "what is a span" at the beginning. And they can go to the spec.md to get all the definitions. I agree with @yurishkuro that we should go for as much consistency as possible. I really like the idea of doing a rev on spec.md next, @azimman. Perhaps we can match the definitions then?

azimman: @pritianka  :+1: I think getting the terminology consistent will be a huge help as the OT community grows.

rpc-ipc.md
```diff
@@ -0,0 +1,249 @@
...
+    new TextMapExtractAdapter(headers));
+```
+
+OpenTracing can throw errors when an extract fails due to no span being present, so make sure to catch the errors that signify there was no injected span and not crash your server. This often just means that the request is coming from a third-party (or untraced) client, and the server should start a new trace.
+
+### Start the span
+
+Once you receive a request and extract any existing span context, you should immediately start a span representing the lifetime of the request to the server. If there is an extracted span context present, then the new server span should be created with a ChildOf reference to the extracted span, signifying the relationship between the client request and server response. If there was no injected span, you'll just start a new span with no references.
+
+Python:
```
yurishkuro: don't exactly know how to do it, but it would be nice to have a style that renders different languages as tabs, so that only one language is seen at a time, making presentation easier

yurishkuro: doesn't have to be this PR

azimman: :+1: ... kind of like the [gRPC docs](http://www.grpc.io/docs/quickstart/cpp.html)

pritianka: Btw, in the quickstart page for the new OT site redesign, I am planning on tabs based info. We only have Go and Python for examples right now but when we have more, we can just add more tabs.

We can copy this here too if it works.

yurishkuro: gRPC site used to have tabs, but now seems they have separate guides for each language (probably auto-generated somehow)

azimman:
Just wanted to gently bump this thread to see if we could merge these docs. I wanted to make sure these were done before submitting another PR for additional updates.

bensigelman:
I basically think this is an improvement and would like to get it in... I may make a few nitpicks, but if there are no great objections I'd propose we merge this in 36h or so.

rpc-ipc.md
```diff
@@ -0,0 +1,249 @@
...
+Before you begin, check [here](http://opentracing.io/integration/) to make sure that there's a working OpenTracing API for your platform.
+
+## Overview
+
+At a high level, here is what you need for an OpenTracing integration:
+
+Server framework requirements:
+
+* Filters, interceptors, middleware, or another way to process inbound requests
+* Active span storage: either a global context or a request-to-span map
```
bensigelman: s/global/request/

rpc-ipc.md
```diff
@@ -0,0 +1,249 @@
...
+
+* The name of the request handler method
+* The name of a web resource
+* The concatenated names of an RPC service and method
+
+### Specifying Which Requests to Trace
+
+Some users may want to trace every request while other may want only specific requests to be traced. You should ideally allow users to set up tracing for either of these scenarios. For example, you could provide @Trace annotations/decorators, in which only annotated handler functions have tracing enabled. You can also provide settings for the user to specify whether they're using these annotations, versus whether they want all requests to be traced automatically.
+
+### Tracing Request Properties
```
bensigelman: We should also suggest some sort of "Span Decorator"... E.g., this thing for GRPC-Go: https://github.com/grpc-ecosystem/grpc-opentracing/blob/master/go/otgrpc/options.go#L30

IMO it could actually obviate this "Tracing Request Properties" thing... let the user decide how to move framework state into the Span. It's not onerous because it's only done once.

rpc-ipc.md
```diff
@@ -0,0 +1,249 @@
...
+    tracer.add_client_span(request, span)
+def process_response(request, response):
+    tracer.finish_client_span(request)
+```
+
+## Closing Remarks
+
+Once you've packaged your implementation, email us at [community@opentracing.io](mailto://community@opentracing.io) with your implementation details (platform, description, github username) and we'll create a repo for you under [opentracing-contrib](https://github.com/opentracing-contrib/), so that others will be able to find and use your integration. You can also find there concrete examples of OpenTracing integrations into different open source projects.
+
+If you're interested in learning more about OpenTracing, join the conversation by joining our [mailing list](http://opentracing.us13.list-manage.com/subscribe?u=180afe03860541dae59e84153&id=19117aa6cd) or [Gitter](https://gitter.im/opentracing/public).
```
bensigelman: nit: is the mailing list URL intended to have the `u=` and `id=` for stuff like this?

yurishkuro:
Please note that without a style change adding more items to the menu will make some menu item unreachable. Right now, on my screen at least, the menu takes all of the vertical space, and it's not scrollable with the rest of the page

image

_config.yml
```diff
@@ -2,19 +2,21 @@
...
   "Semantic Specification": '/spec'
   "Contributors": '/contributors'
   "News and Notes": '/updates'
 
-user_pages_list:       
+user_pages_list:
   "APIs and Supported Tracers": '/integration'
   "Instrumentation Use Cases": '/use-cases'
   "Cross-Process Propagation": '/propagation'
+  "Instrument Systems at Scale": '/systems-at-scale'
```
azimman: 👍

_config.yml
```diff
@@ -2,19 +2,21 @@
...
 
-implementor_pages_list:       
+implementor_pages_list:
   "Implementing OpenTracing": '/implementors'
+  "RPC/IPC Frameworks": '/rpc-ipc'
```
azimman: 👍

rpc-ipc.md
azimman: updated.

rpc-ipc.md
azimman: added in the Span Decorator example.... left in the Tracing request priorities, but if you have a preference for removing it I'm not attached to it.

rpc-ipc.md
azimman: ... nope. Removed.

pritianka:
Hi Yuri - I am working with a designer to redo our website (as I shared in
the OT Gitter) so that should take care of it :-)

Priyanka Sharma
Co-founder, WakaTime <http://www.wakatime.com/> - Quantify your coding
<http://www.twitter.com/pritianka>
www.twitter.com/pritianka
650-796-7125

azimman:
@yurishkuro what browser are you using? 

image

RFOMM (Renders fine on my machine... 😕 )

bensigelman:
@azimman it's not a browser issue, it's a window-height issue... I think @yurishkuro's point is that the leftnav doesn't scroll. @pritianka's redesign comment is certainly relevant, but we don't want the leftnav to be busted in the interim. I can try to fix this with css or something.

azimman:
@bhs got it. I'll take a look tomorrow at the current template and see if there is an easy fix.

azimman:
@yurishkuro can you take a look at the latest commit and see if it works for your default window size. This is how is renders for a 1280x800 resolution: 

image

Easy to reduce it further if desired... but will obviously start to look small at higher resolutions. 

@pritianka It would also be good to provide the designers with minimum resolution and browser requirements as part of the redesign (happy to help you with this).

yurishkuro:
@azimman based on your screenshot looks like it will be ok.

azimman:
Any other updates needed before merging this?

azimman:
added new docs to site redesign.

PR #120, opened 2016-09-28
Fix typo on use_cases page

trentonsmith-wf:
`and one o them is the so-called "sampling priority"` -> `and one of them is the so-called "sampling priority"`


bensigelman:
Gracias!

PR #122, opened 2016-10-09
Redesign the OpenTracing website

reimertz:


bensigelman:
Since we need this to be live before tomorrow morning, I'm going to merge now and check links, etc, etc. I have a bunch of little tweaks I want to make as well, but easier to do that post-merge as well.

@reimertz *thank you*!

Issue #127, opened 2016-10-12
The build doesn't work for me


yurishkuro:
First, `npm run dev` fails when the default Node installation is 0.10, we should specify in the readme which minimal version is required, or better yet add `.nvmrc` file so that people can just  run `nvm use`.

Once I got it to compile, the server at http://127.0.0.1:4000/ just shows the directory.  I had to rename the file `$ mv _site/index _site/index.html` to show the home page.

cc @reimertz

reimertz:
@yurishkuro PR above should solve your issues. Please pull it and confirm, if see, let's push it to master.

also, thanks for helping me sorting out all these small things. 👍

reimertz:
@yurishkuro can we close this issue now?

yurishkuro:
ok, thanks

PR #130, opened 2016-10-13
add .nvmrc with  node v4.6.0 (latest TSL), update readme

reimertz:
this pull-request should solve the issues surfaced in ticket #127.

cc @yurishkuro @bensigelman


.nvmrc
```diff
@@ -0,0 +1 @@
+4.6.0
```
reimertz: I just wanted to go with the latest version that is under [TLS](https://github.com/nodejs/LTS#lts-schedule).

4.6.0 includes the latest security patches, but if you prefer v4.0 or if I am missing a major perk of going with 4.0, please remind me. :)

Issue #128, opened 2016-10-12
Not clear how to submit PRs


yurishkuro:
Aside from changing the actual source file, does a PR need to include the files that get modified under `documentation/` folder? I haven't even changed anything in the documentation, only the homepage. Please clarify in the README.

@reimertz

bensigelman:
@yurishkuro agree this should be clarified... I got by with a one-time `npm install` followed by `npm run production` prior to submit.

There are a bunch of genfiles which change every time and which should probably be part of a TravisCI step. @reimertz would be better able to make specific suggestions.

yurishkuro:
Or even better not to have those files changes. As far as I can tell, the change is two-fold, some timestamp is included in JS inside HTML, not sure why, and a couple of js files get a filename based on unix timestamp (things like webpack usually just hash the file content for cache busting, which results in stable file names).

Overall, I am not sure why we need Node to be involved in a jekyll-based site, but then again it was clearly a lot of work to renovate the website that I personally didn't have to do, so I won't complain :-).

bensigelman:
@yurishkuro I think the TL;DR is `gitbook` which has some fancy client-side search features, etc. But I don't actually know (or, frankly, care) either.

Usually when timestamps are involved, it's a way of guaranteeing that browser caches get cleared "no matter what". Not sure if that's the case here.

reimertz:
@yurishkuro I agree with you, the process right is too complex right now and the reason is, as @bensigelman mentioned, because we use gitbook for the documentation part of the page.  I think this flow ended up a bit backwards and needs to be solved.

I think the best solution for this is to add /documentation in the .gitignore and instead run build steps with travis once we get a push to master.

We can then add a badge to in the readme to will show if the latest commit on master is broken or not. 

To clarify how a PR should be done, I could create CONTRIBUTE.md with steps necessary to submit a PR.

PR #131, opened 2016-10-13
add travis build

reimertz:
partly a fix for #128

this commit adds automatic build for out gh-pages branch.

what this means is that we don't need to run `npm run production` anymore. what is nice about this is that we don't get bloated commits anymore since /css and /documentation are excluded from the commits.

furthermore, it means that changes to /_docs will be built automatic, so if people prefer to do the documentation edits over here at github, you are free to do so.

what is lacking in this commit is a proper CONTRIBUTE.md file. I would love if someone with a bit more knowledge could chip in here.

Would also love if someone could pull this pr and try it.

one thing that needs to be done before this will work is to add an access token to the travis build once we merge this commit.

It is pretty straight forward.
1. got to https://github.com/settings/tokens/new 
2. create a token with repo and user permissions
3. copy-paste the token generated by github
4. go to https://travis-ci.org/opentracing/opentracing.github.io/settings
5. add Environment variable named ACCESS_TOKEN, with the newly generated token as a value. 
6. CAUTION: DO NOT CHECK "Display this value in build log". 
7. Everything should now work. :)

cc @bensigelman @yurishkuro


reimertz:
( I cancelled the build since it would push this to gh-pages. :) )

README.md
```diff
@@ -19,34 +28,16 @@ jekyll --version
 ## Running Opentracing.io locally
 
 ### Install dependencies
-
 ```
 npm install
```
yurishkuro:
```
nvm use
npm install
```

PR #133, opened 2016-10-16
Add Sourcegraph to Organizations using OpenTracing

slimsag:
screenshot:
image

We're using OT pretty extensively throughout our system now and are loving it, so we'd love to be an advocate :)


bensigelman:
@slimsag excellent, thank you! I should have thought of this myself.

One quick question: how does this render on mobile (or in Chrome's mobile sim)? I'm thinking that the list may be a bit too wide at this point... hm.

slimsag:
Good question! I hadn't thought of that at all. I just tried and on larger devices it looks okay-ish, but smaller ones seem to horizontally stretch the images inward:

image

In fact, it happens a little bit (although certainly not as much) without this change today (look at the Joyent icon):

image

I assume we should just break it up into two separate rows or maybe something else; I'm not 100% sure how to solve this in the CSS, but perhaps I can try giving it a shot later next weekend or so.

bensigelman:
@reimertz @fractaly have any quick fix in mind here?

reimertz:
Maybe we could  wrap to a new line?

yurishkuro:
If we have a grid system in whichever css we're using, one way to do this is to put each icon in a cell and give the cell different proportions depending on the screen size. E.g. in bootstrap it would be something like `"col-lg-2 col-med-3 col-sm-6 col-xs-12", which means the icon would take 2/12th on large screen, 3/12th on med, etc, and full row on x-small screen.

The only catch here is that each icon gets equal width, which may look great if icons of different width.

But simply allowing wrapping of the line might be the simplest.

pritianka:
I expect as more people want to put their logos on (for companies as well as frameworks and maybe tracers some day), the two lines will stop being a solution too. Hence I had a separate page for company logos in initial spec. We could do one of three things - 

1. Multiple lines for a logo filled homepage (wont work for the frameworks and tracer if people start adding logos there). 

2. People click on a "more" after the first line for more logo lines to show more logos on the same page. 

3. People click on a "more" link to be taken to a company logos page. 

Priyanka

azimman:
another option could be a rotating gif could even have 2-4 logos per frame.

bensigelman:
Most landing pages for standards have a wall of logos somewhere, and I am ok moving in that direction... For now my preference would be to wrap to two lines as screen dimensions warrant...

yurishkuro:
+1 for companies/tracers/frameworks landing pages, can't fit everything on the home page.

-1 for animated git, they are a pain to maintain. A JS-based slider might be a better option. Or just wait until we actually have so many logos to warrant it.

bensigelman:
I'm a `-1` for a separate "companies" page for the time being. All we can really say is that OT is being used at a given company... there won't be details beyond that, at least until we have proper OT "case studies" (which feels like a long way off). I can see a separate "frameworks" page, though, and of course we already have a (very limited!) "tracers" page.

To move this forward and hopefully merge things quickly: can we have the css wrap these in groups? I.e., do something so that if we need to wrap, we don't wrap just a single "straggler" logo/icon.

reimertz:
I just created a PR that will make this look better on mobile and smaller devices.

image

bensigelman:
@reimertz thanks!

bensigelman:
@slimsag can you rebase this and verify that it still works, esp for mobile screens? If it looks good I think we can merge now. Thanks!

bensigelman:
(@slimsag friendly ping!)

slimsag:
@bensigelman so sorry for the delays here. I got caught up in a few other things that distracted me from this. I will update by EOD.

slimsag:
Rebased on to latest `master`, updated screenshots for desktop/mobile:

Desktop:

image

iPhone 5:

image

iPhone 6:

image

Sorry for my delays here, again, and thanks for the patience :)

PR #140, opened 2016-10-23
Defer to the CNCF code of conduct

bensigelman:


bensigelman:
I'm assuming that this will be uncontroversial, but I'll leave it up for a few days just in case there are concerns. If not, I will have the per-language repos link to this (or maybe to CNCF directly).

bensigelman:
Going once, going twice... I will merge this tomorrow sometime with no objections. Thanks to those who commented on Gitter, too.

PR #141, opened 2016-10-23
make companies logos wrap when there isn't enough space

reimertz:
fixes issues mentioned in  #133


bensigelman:
Do you need to `npm publish`?

reimertz:
There are no changes that this affect on the current state of the branch so I though it was unnecessary to do it. But it will be needed when we want to deploy the new logo. :)

bensigelman:
As a best practice, I think we ought to mandate an`npm publish` with every merge. It will make diffs more predictable until we have a better solution.

Also, I think on mobile we already have too many logos for one line (even without the extra one).

reimertz:
@bensigelman okay, i ran the production script now and pushed a new commit containing the changes.

bensigelman:
thank you, @reimertz!

PR #145, opened 2016-10-27
linkedHeadings: linked headings to ids

dustinlessard-wf:
## Problem

We often want to link to specific sections in the spec, but cannot do this easily because the headings aren't linked as anchors.
For example, we can go to http://opentracing.io/documentation/pages/spec#spans directly, but discovering this link exists is not straightforward.  I'd expect to click on this heading to change the url to one including the anchor id.
http://screencast.com/t/hL6HchXXO

## Solution

Rather than edit every occurrence of h2 and h3 tags, the approach taken in this change is to simply auto-wrap these in anchor tags via javascript.  This will apply to all h2 or h3 tags, regardless of page.
http://screencast.com/t/UlCpHo5TM

### DIFF
--- a/_docs/scripts/linkedHeadings.js
+++ b/_docs/scripts/linkedHeadings.js
@@ -1,3 +1,3 @@
-$('h2, h3').filter('[id]').each(function () {
+$('h1, h2, h3, h4').filter('[id]').each(function () {
     $(this).html('<a href="#'+$(this).attr('id')+'">' + $(this).text() + '</a>');
 });
--- a/_docs/scripts/linkedHeadings.js
+++ b/_docs/scripts/linkedHeadings.js
@@ -1,0 +1,3 @@
+$('h2, h3').filter('[id]').each(function () {
+    $(this).html('<a href="#'+$(this).attr('id')+'">' + $(this).text() + '</a>');
+});
--- a/book.json
+++ b/book.json
@@ -11,7 +11,8 @@
      "scripts": {
         "files": [
           "scripts/injectLogo.js",
-          "scripts/analytics.js"
+          "scripts/analytics.js",
+          "scripts/linkedHeadings.js"
         ]
     },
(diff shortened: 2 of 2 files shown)


_docs/scripts/linkedHeadings.js
```diff
@@ -0,0 +1,3 @@
+$('h2, h3').filter('[id]').each(function () {
```
bensigelman: worth doing for `h1` and `h4`??

bensigelman:
(cc @reimertz)

_docs/scripts/linkedHeadings.js
dustinlessard-wf: Currently, the cursor does become a pointer despite the screencasts attempts to hide that.
tags

For consistency, we could make the `h1` and `h4` also clickable - will do.

dustinlessard-wf:
Currently, the cursor does become a pointer despite the screencasts attempts to hide that.
tags

bensigelman:
LGTM, thanks. FYI, this will not go live until we do an `npm publish` and push. Probably will be in the next 24h or so (@reimertz was going to change some impl details about the push process to make the `npm publish` part of TravisCI).

dustinlessard-wf:
Sounds good - thanks guys!

reimertz:
agree and good work @dustinlessard-wf!

Just curios, did you check if there was a plugin doing this already at https://plugins.gitbook.com/?

Anyway, I will test and initiate the migration over to Travis CI tomorrow.

dustinlessard-wf:
Sorry guys, I didn't check for a plugin.   I found a couple that seem to do the same thing.  If you prefer, I'd be happy to pr a change to using a plugin?

reimertz:
I think that might be a better option, because when I test this now, it seems it only works for the first load, but then, when you navigate further, content is fetched async -> new headers wont be links.

PR #147, opened 2016-10-28
this commit should fail

reimertz:
added a include that doesn't exist just to make sure that travis catch errors


reimertz:
yep, travis is acting appropriate here, not deploying and also triggering a failed build.

PR #156, opened 2016-11-07
Add Hawkular to the list of supported tracers

bensigelman:
cc: @objectiser -- thanks for doing the work on your end!

### DIFF
--- a/_docs/pages/supported-tracers.md
+++ b/_docs/pages/supported-tracers.md
@@ -33,3 +33,7 @@ For more details, read [the godocs](https://godoc.org/github.com/sourcegraph/app
 ## LightStep
 
 [LightStep](http://lightstep.com/) runs a private beta with OpenTracing-native tracers in production environments. There are OpenTracing-compatible [LightStep Tracers](https://github.com/lightstep) available for Go, Python, Javasrcipt, Objective-C, Java, PHP, Ruby, and C++.
+
+## Hawkular
+
+[Hawkular APM](http://www.hawkular.org/hawkular-apm/) supports OpenTracing-Java and has plans to support other platforms in the near future.
--- a/_data/tracers.yaml
+++ b/_data/tracers.yaml
@@ -19,3 +19,7 @@ array:
   - name: Jaeger
     url: http://uber.github.io/jaeger/
     icon: false
+
+  - name: Hawkular
+    url: http://www.hawkular.org/hawkular-apm/
+    icon: false
--- a/_includes/hero.html
+++ b/_includes/hero.html
@@ -27,7 +27,7 @@
             </a>
         {% endfor %}
       </div>
-      <div class="logo-bar__logos frameworks">
+      <a class="logo-bar__logos frameworks" href="/documentation/pages/supported-tracers">
         {% for framework in site.data.frameworks.array %}
           <a href="{{framework.url}}">
             {% if framework.icon %}
@@ -37,7 +37,7 @@
             {% endif %}
           </a>
         {% endfor %}
-      </div>
+      </a>
     </div>
   </section>
 </section>
--- a/_includes/hero.html
+++ b/_includes/hero.html
@@ -27,7 +27,7 @@
             </a>
         {% endfor %}
       </div>
-      <a class="logo-bar__logos frameworks" href="/documentation/pages/supported-tracers">
+      <div class="logo-bar__logos frameworks">
         {% for framework in site.data.frameworks.array %}
           <a href="{{framework.url}}">
             {% if framework.icon %}
@@ -37,7 +37,7 @@
             {% endif %}
           </a>
         {% endfor %}
-      </a>
+      </div>
     </div>
   </section>
 </section>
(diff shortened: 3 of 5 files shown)


_docs/pages/supported-tracers.md
```diff
@@ -33,3 +33,7 @@ For more details, read [the godocs](https://godoc.org/github.com/sourcegraph/app
 ## LightStep
```
yurishkuro: We should link to this page from the section header on the landing page

yurishkuro:
LGTM, but has a conflict

pritianka:
Hi @objectiser - Thanks for your work on this! Would you be open to chatting about Hawkular for the OpenTracing blog? An interview-style post would make most sense but am happy to do a different format if you prefer.

_docs/pages/supported-tracers.md
bensigelman: @yurishkuro do you want Jaeger to be its own subsection, independent of Zipkin?

bensigelman:
This ended up being non-trivial css-wise for two reasons:

1. The number of tracer impls is large enough that they don't all fit at many screen resolutions
1. @yurishkuro it's actually pretty tricky to link from the "Supported Tracers" thing because it's part of a `:before` thing, not normal doc contents (i.e., it's in the css, not the html).

@reimertz would you be willing to adopt this branch and fix the problems above? Thanks!

pritianka:
Since the list of tracers (and hopefully the projects instrumented for OT
as well soon) is increasing, I'd like to revisit the idea of a separate
page that we can link to from the homepage after we've displayed the max
number of logos possible there. This page can have some detail around each
tracer and links to documentation/repos.


ᐧ

Priyanka Sharma
Advisor, HeavyBit <http://heavybit.com/>
Head of Marketing & Partnerships, LightStep <http://lightstep.com/>
Co-founder, WakaTime <http://www.wakatime.com/> - Quantify your coding
www.twitter.com/pritianka
650-796-7125

bensigelman:
@pritianka that page already exists, actually :)

http://opentracing.io/documentation/pages/supported-tracers

Maybe we should choose the ~4-5 that are most actively developed for the homepage... seems like a fine solution to me. Thoughts from others? (@yurishkuro?)

pritianka:
Oh right - sorry, I confused this with framework and companies pages that we didn't do last time. 

The word tracers on the homepage can link to this page and we can have all the tracers listed here and only the most active ones can be on the homepage?

yurishkuro:
"Most active" is hard to measure objectively. I'd rather list all until this proves to be a problem.

objectiser:
@bensigelman Would you like me to provide some text for the supported tracers page aswell?

@pritianka Yes would be interested - could you contact me at gbrown@redhat.com so we can discuss details.

pritianka:
Sent you a note @objectiser. Talk to you soon!

wu-sheng:
I think we should create a new section of Index Page, like `Why OpenTracing` section, and just below it. And we can layout many supported Tracers and supported Frameworks. Like a **OpenTracing Supporter Wall**.

reimertz:
@benaigelman what do you think about inviting @fractaly to this discussion, maybe she could help us figuring out how to fit n supporters / frameworks on the site with a new design if we prefer to show all of them on the landing page? 

I see both pros and cons of this, a supporters list could get very long, to the point where it will be messy to show all of them, but on the other hand could be solved with some semi-fancy carousel etc. 

Once we have a firm decision on how this should be solved,  I'll start working on it asap. :)

bensigelman:
@reimertz sure, I'd be glad to have @fractaly's help! And a carousel is a fine way to deal with the problem. For a while there won't be an overwhelming number of these implementations... probably no more than 10 for the next year or so, I'd guess/hope?

fractaly:
Yeah, I agree a carousel makes sense. Especially when you get those first new tracers/frameworks that creates a new line, but it doesn't really look good to add another line yet.. I also think it would make sense to make the background divider be centered, to get an even amount of space for both tracers and frameworks. That will probably open up for about 6 logos in each section on a standard desktop view, and maybe 4-5 on mobile.

bensigelman:
Alright. I'm going to do something "pragmatic" to list Hawkular. Once there's a functional carousel, it will look better :)

wu-sheng:
Looking forward the `carousel`. Hope I can catch up to add sky-walking on it.     :) :P

Too busy these two weeks.

bensigelman:
@wu-sheng sounds good!

bensigelman:
I removed Hawkular from the OT landing page until we have a carousel to use there. I kept the changes to the "supported tracers" page, of course. Merging this PR for the time being...

wu-sheng:
@bensigelman `HAWKULAR` already in landing page. But it cause the page very *ugly*

It is like a char seq, not several tracer icons or names.

qq20161117-0

bensigelman:
@wu-sheng argh, thank you, yes... I neglected to push a local change before merging. Fixed in master.

reimertz:
Will start implementing the carousel, but push a quick fix to make stuff look good with one more tracer. :)

PR #157, opened 2016-11-10
fix incorrect section name

wu-sheng:


yurishkuro:
Thanks, @wu-sheng

wu-sheng:
I am going to read and translate the doc. Send rp if I find any other issues.

PR #159, opened 2016-11-14
Add myself as a contributor

wu-sheng:
@bensigelman , I have add myself on the contributors page.


bensigelman:
@wu-sheng the actual change looks good... can you perhaps squash these commits down in `git`? (There are a bunch of commits that don't pertain to the PR here)

yurishkuro:
@bensigelman squashing can be done during the merge

image

We could change the repo setting to disallow full merge altogether, but for some reason right now GitHub doesn't let me change it

image

wu-sheng:
I will send a new RP instead. @bensigelman , please check what @yurishkuro mentioned.

wu-sheng:
Already send #160.

PR #160, opened 2016-11-14
add myself as a contributor

wu-sheng:
send this rp again, same with #159, only one commit log. :) 

@bensigelman @yurishkuro


reimertz:
@bensigelman @wu-sheng @yurishkuro

What I think we could do is to use this: https://help.github.com/articles/repository-metadata-on-github-pages/

and make contributors list dynamic. Should be a slamdunk.

bensigelman:
@reimertz unfortunately it's harder with OpenTracing since most of the contributions are in the non-docs repos. :(

Yuri had previously pointed to a javascript-based solution to this problem that's more flexible. Yet another idea would be to make the update happen as part of the TravisCI step for the docs repo, though not sure who's going to volunteer to do that work :)

PR #165, opened 2016-11-25
Added Instana tracers to homepage and docs. 

CodingFabian:
Tweaked css to accommodate the large number of tracers.

New look:
screen shot 2016-11-25 at 20 35 34

screen shot 2016-11-25 at 20 43 45

### DIFF
--- a/_docs/pages/supported-tracers.md
+++ b/_docs/pages/supported-tracers.md
@@ -32,8 +32,12 @@ For more details, read [the godocs](https://godoc.org/github.com/sourcegraph/app
 
 ## LightStep
 
-[LightStep](http://lightstep.com/) runs a private beta with OpenTracing-native tracers in production environments. There are OpenTracing-compatible [LightStep Tracers](https://github.com/lightstep) available for Go, Python, Javasrcipt, Objective-C, Java, PHP, Ruby, and C++.
+[LightStep](http://lightstep.com/) runs a private beta with OpenTracing-native tracers in production environments. There are OpenTracing-compatible [LightStep Tracers](https://github.com/lightstep) available for Go, Python, Javascript, Objective-C, Java, PHP, Ruby, and C++.
 
 ## Hawkular
 
 [Hawkular APM](http://www.hawkular.org/hawkular-apm/) supports OpenTracing-Java and has plans to support other platforms in the near future.
+
+## Instana
+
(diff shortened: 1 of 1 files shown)


bensigelman:
@CodingFabian I don't like this from a design perspective... but I don't think you should have to fix it, eithier :)

@reimertz where are we with the carousel impl? there are now ≥2 projects waiting on it and it seems (?) like it should be relatively easy. Thanks.

yurishkuro:
There's a weird effect that each subsequent "Tracer" implementation gets bigger and bigger font. For example, both TRACER and JAEGER use plain text names, not images, yet JAEGER looks bigger. And HAWKULAR is even bigger that those two.

CodingFabian:
@bensigelman well agreed, as said its a proposal. I was inspired by the max-width 780px design which does the same wrapping when you resize the browser. i thought now that there are more logos why not always wrap it.
screen shot 2016-11-25 at 22 02 12
Thats the current page when you resize it.

@yurishkuro you are right. i tried to fix that manually, but it seems that this has always been the case. see above screenshot of the current version.

I tried to fix both issues, but apparently my fix is not good enough.
if the layout on the homepage is a showstopper, then i can cherry pick at least the documentation part into a separate pr.

CodingFabian:
if we want to accept this styling also for wider screens, then maybe @yurishkuro and I can try to tweak this pr to make the font sizes align.

bensigelman:
@CodingFabian the css cut-point for "medium-sized screens" is aimed at tablets.

If you'd like to restrict this PR to the `.md` file for "supported tracers", I'll happily merge now. I would like to spend the like 48h (or whatever) to keep the intended design intact via a carousel, though, as there is nothing super urgent about the change to the homepage (at least as far as I understand – correct me if I'm wrong about that).

wu-sheng:
@bensigelman , should we wait for the **carousel** impl? This homepage style is not look good for me. The `supported tracers and frameworks` hold too much areas in the page, this section should be a additional info, not the major info.

bensigelman:
@wu-sheng that is what I'm suggesting, yes (that we wait for the carousel implementation). My understanding is that it should be ready before Monday.

wu-sheng:
@bensigelman , wonderful, looking forward!

CodingFabian:
I am traveling at the moment, please feel free to cherry pick from this PR. otherwise i will make new ones next week.

reimertz:
@bensigelman @wu-sheng  Yep, working on the carousel as we speak, will have a PR by EOD.

CodingFabian:
Now that the carousal is tackled by a separate script and #166 includes our logo and link. this pr is just merely the documentation. I guess it can be merged without further changes?

bensigelman:
@CodingFabian thanks for your contribution and for your patience!

PR #166, opened 2016-11-27
Add responsive carousel that support n tracers/frameworks

reimertz:
finally, here is the carousel implementation. 

**it is:** 
- responsive
- supports n tracers/frameworks
- stops rotating on hover
- implemented the logic myself. felt unnecessary to pull in jQuery + plugin (~300kb) for one feature. 

**demo**

opentracing_carousel

cc @bensigelman


wu-sheng:
@reimertz , It's a nice work. I like it.

_data/tracers.yaml
```diff
@@ -20,8 +20,10 @@ array:
...
-#  - name: Hawkular
-#    url: http://www.hawkular.org/hawkular-apm/
-#    icon: false
+  - name: Hawkular
+    url: http://www.hawkular.org/hawkular-apm/
+    icon: false
+
+  - name: Instana
+    url: https://www.instana.com
+    icon: https://www.instana.com/wp-content/uploads/2016/07/header-logo.svg
```
yurishkuro: I would recommend copying the logo into this repository. The external link might get broken or the logo dimensions might change and break the layout.

reimertz: Totally agree. 

@CodingFabian, would it be okay for us to copy-paste the image referred here to this repo?

_sass/_components/_logo-bar.scss
```diff
@@ -86,6 +65,8 @@ $logo-bar-scale: 14px;
       text-decoration: none;
       font-size: 1rem;
       justify-content: center;
+      min-width: 25%;
+      transition: margin-left 750ms;
 
       &[href*="tracer"],
       &[href*="dropwizard"] {
```
yurishkuro: Q: as I understand, both "tracer" and "dropwizard" are included with `icon: false`, i.e. as plain text. Is it possible to change the generation script to use a generic name in this case, like "plain-text-logo", so that when new participants are added we do not have to keep changing the css?

reimertz: Yeah, you are right. 

The issue is that the initial implementation assumed we would have logos for all tracers/frameworks.  So what I am doing here are small tweaks to make the text and logos align a bit more beautiful. 

A iteration on this implementation would solve these issues for sure.

bensigelman:
@reimertz my only feature request is that the "supported tracers" text itself is clickable and links to the dedicated page describing these various implementations. I was going to do it myself but it's not entirely trivial since that text is part of the CSS, not the HTML.

Thanks!

_sass/_components/_logo-bar.scss
```diff
@@ -101,11 +82,17 @@ $logo-bar-scale: 14px;
           max-height: $logo-bar-scale * 1.25;
         }
 
-        &[src*="zipkin_small.png"] {
+        &[src*="grpc"] {
```
bensigelman: I am not enthusiastic about putting yaml-like content in here... is there an easy alternative to this approach?

reimertz: @bensigelman making 'supported tracers' clickable + fixing this would go well together. My initial implementation assumed images, but the introduction of text-only made things a little bit...quirky. 

So some css changes + turning the pseudo-classes into actual html is my proposition here. Should be fairly quick. :smile:

bensigelman: Sounds good, @reimertz. Would you like to make these changes as a separate PR? I'm fine either way.

reimertz: @bensigelman Yeah, let's do that.

reimertz:
@CodingFabian thanks for your other commit to add Instana. 👍 

Just wonder if it would be okay for us to copy-paste [your logo](https://www.instana.com/wp-content/uploads/2016/07/header-logo.svg)?

CodingFabian:
@reimertz sure!

CodingFabian:
also feel free to resize the image if needed

reimertz:
@CodingFabian image added. 
@bensigelman i took the freedom to add lightsteps logo as well, I hope this was okay? If so, I'll merge this into master.

bensigelman:
@reimertz thanks!

wu-sheng:
@reimertz thanks a lot.

PR #170, opened 2016-12-01
Use Hawkular Logo Text (white)

theute:
Sorry, I couldn't test, but this should replace HAWKULAR with our logo (without the image)

Thanks for adding Hawkular on the homepage !


jpkrohling:
This is how it looks like to me: https://snag.gy/kHoEuS.jpg

bensigelman:
Thanks!

PR #171, opened 2016-12-01
fix ios render bug

reimertz:
Tracers and frameworks on certain viewports got a weird render bug that
caused both frameworks and tracers to render on the same row as Tracers.


bensigelman:
LGTM, thanks!

PR #175, opened 2016-12-21
Add motan rpc to supported framework.

wu-sheng:
Only add a framework.


wu-sheng:
@reimertz ,  Am I do the right thing to add a supported framework?
@bensigelman , This is the most popular open-source active rpc framework in China, which belongs to Sina Weibo, something is similar to twitter.  I am happy to invite them to join the OT. 🎉🎉🎉🎉

bensigelman:
Thanks! This is the essence of it, right? https://github.com/weibocom/motan/blob/master/motan-extension/filter-extension/filter-opentracing/src/main/java/com/weibo/api/motan/filter/opentracing/OpenTracingFilter.java

wu-sheng:
Yes, it is.  

The other rpc framework's java edition will be ready soon.😃
It's the happy week.

PR #180, opened 2017-01-04
Fixed a typo

joonas:
Noticed this while reading through the docs so I thought I'd open a PR.


wu-sheng:
Definitely a typo. Thank you.

PR #183, opened 2017-01-22
Change translation page

wu-sheng:
Add spec-zh translation link. Spec and online document(http://opentracing.io/documentation/) have different purposes, and online document said more thing about backgroud, scenario and use-cases. I want to keep both of them right now. Hope it's OK with you. @bensigelman 

wx20170122-175314 2x


wu-sheng:
@bensigelman , should I merge this? or, need to change something?

bensigelman:
Let's merge it -- thank you again!

wu-sheng:
Merged. Do what I can do best :)

PR #190, opened 2017-03-17
Add hprose in supported frameworks.

wu-sheng:
Finally released. And here is the snapshot:

wx20170317-174655


wu-sheng:
btw @andot @hprose

andot:
Thank you very much @wu-sheng

wu-sheng:
@bhs Can we merge this?

bhs:
sure! sorry for the delay.

PR #194, opened 2017-03-24
Show all videos in the videos section. 

reimertz:
local build differed from github pages build because Github has updated their dependencies to use 3.4.3 (https://pages.github.com/versions/) which changed how liquid templates handles the case when a variable is non-existing.

This change resulted in condition `  array | where "featured", ""` not getting fullfilled when `featured` was not set.

Fixed this by adding feature variable to all videos. :)


bhs:
@yurishkuro that changed a few months ago IIRC

PR #199, opened 2017-04-06
Added CloudNativeCon videos.

jpkrohling:


jpkrohling:
@pritianka , @reimertz , could you check if that's all it's required to include the videos?

reimertz:
Thanks for this PR!  👏

I just verified it locally and made sure copy is same as youtube description.

I think we're ready to merge this.

PR #203, opened 2017-05-09
Defer in separate functions

bhs:
... and call .Close()


bhs:
IRL :thumbsup: from @yurishkuro

PR #207, opened 2017-05-30
Restructure how we serve fonts and updated bariol to a working version.

### LINKED ISSUE #201: Fonts are looking broken across all browsers
FF v53, Chrome v58 on Windows, IE 11, Edge 38 - fonts on the site look broken:
image
Notice jumping size of non-capitalized letters. It seems to be an issue with the bariol font. In FF I can see a request to the font `http://opentracing.io/documentation/styles/bariol_regular-webfont.eot` which returns 404. Disabling this font using development tools fixes the problem

reimertz:
The one issue we had was that certain browsers didn't like data-uris of the font and used the fallback instead.

Weirdly, the bariol font itself was malformed which resulted in "downloadable font: rejected by sanitizer" errors.

Replaced the font with a updated version and errors went away. 😐

Also removed the inlined data-uris to decrease the size initial css being sent over the wire.

Important change can be seen here: https://github.com/opentracing/opentracing.io/compare/master...reimertz:fix-font-errors?expand=1#diff-5486396041702fd41bc406ab7f004be1R2

fixes #201


bhs:
thank you @reimertz !

PR #221, opened 2017-08-29
Add deploy description

reimertz:
@tedsuo had a valid point of it being unclear how we deploy changes once a PR is merged so here is a short description for future contributors. :)


reimertz:
@tedsuo updated the description

PR #230, opened 2017-12-07
Add sharding-jdbc as supported framework.

wu-sheng:
Sharding-JDBC is a JDBC extension. Sharding is a common way in China, when too much data for a single MySQL, but NoSQL or some other newSQL Database, because of Ops-concern or traditional Transaction requirements. It support to route(sometimes change) sql to different MySQL instances(or table) by hash, range or other rules, according to analysis result of SQL statement at run time.

I worked with their team on this, and tested by using SkyWalking OpenTracing Bridge with [sharding-jdbc-opentracing](https://github.com/shardingjdbc/sharding-jdbc-opentracing)

Thanks for @hanahmily @terrymanu .


yurishkuro:
could you post a screenshot how it looks in the UI?

wu-sheng:
It looks like this:

image

wu-sheng:
And the CI error is:
```
ERROR:  Error installing jekyll:
	ruby_dep requires Ruby version >= 2.2.5, ~> 2.2.
```

I assume It is not triggered by my chances, right?

yurishkuro:
Thanks. 

The error is not sure to your changes, but your changes may not make it to the website because of it, so we should fix it first. Looks like a dependency got stale, but I'm not seeing where is defined.

wu-sheng:
@reimertz Maybe you can help us on this, please.

wu-sheng:
@tedsuo @bhs @yurishkuro Anyone can help me on this CI issue? Seems I am being blocked.

tedsuo:
@wu-sheng let me know if a rebase does not solve this.

PR #242, opened 2018-02-23
Updated Outreachy page for the new program

jpkrohling:
Signed-off-by: Juraci Paixão Kröhling <juraci@kroehling.de>

### DIFF
--- a/pages/outreachy-201702.md
+++ b/pages/outreachy-201702.md
@@ -1,0 +1,193 @@
+---
+permalink: /outreachy.html
+layout: page-containing-markup
+title: 'Outreachy'
+---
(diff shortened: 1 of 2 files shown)


jpkrohling:
@jstickler, would you please review?

@pritianka, @reimertz, could you also take a look? If it looks OK to you and Julie, would you please merge?

JStickler:
Looks good to me, my only question is do we want to explicitly say that we're looking for an intern to work on documentation for this round?

jpkrohling:
I think so. We can then use that on our tweets, to get special attention from people looking for non-technical tasks.

PR #250, opened 2018-03-15
Updated OTOV_3.png

### LINKED ISSUE #231: Typo in OTOV_3.png: 'stoage' -> 'storage'
I cannot find the sources to fix that, but there is a typo in documentation/images/OTOV_3.png: instead of 'storage allocation' it says 'stoAge allocation'.

AarshiyaGuneja:
Fixes #231


yurishkuro:
what is the source of this png? any chance we can store the original, not just the rendered version?

also, the new image seems to be large, is that going to mess up the site layout?

AarshiyaGuneja:
I have made this image myself using Lucidchart.

AarshiyaGuneja:
I have resized the image as original so that it doesn't create any issues further.

yurishkuro:
is it possible to expose the original source of the image from lucidchart? I case we need to change it in the future would be nice to have the vector based original.

AarshiyaGuneja:
Yes, it can be done. 
I have made the changes to the original image as well. I'm attaching the image here. If it seems appropriate to you, I'll commit the changes.
otov_3

yurishkuro:
> I have made the changes to the original image as well. I'm attaching the image here.

To clarify, did you work on the actual rendered image, or some vector objects (like shapes in google docs)? My question was about the latter, since you mentioned lucidcharts.

AarshiyaGuneja:
I have used some shapes to design the image as original in Lucidchart.

Issue #252, opened 2018-03-16
Add LICENSE, setup & running information to README


PikBot:
Add the following to README
- LICENSE information
- Setup of the opentracing.io project
- Running commands

reimertz:
Hi there @PikBot and thanks for helping us maintaining this project! :)

I would say that the most common way of communicating license information is to include a [LICENSE](https://github.com/opentracing/opentracing.io/blob/master/LICENSE.md) file in the root of the project which we currently do. :)

Regarding project setup and how to get the environment to run locally, we currently have all the documentation in [CONTRIBUTING](https://github.com/opentracing/opentracing.io/blob/master/CONTRIBUTING.md). 

But once the site is migrated over to Hugo, a lot of documentation will change, so let's keep this issue as a remainder.

PikBot:
Hey @reimertz, sounds good to me! :)

yurishkuro:
I think we already have all of this.

PR #253, opened 2018-03-17
For #252: Add License information to README

shresths:
Addresses one point of #252, for License information on the README file.


README.md
```diff
@@ -11,3 +11,9 @@ Contributions are always welcome, no matter how large or small. Before contribut
 please read the [code of conduct](code-of-conduct.md).
 
 See [Contributing](CONTRIBUTING.md).
+
+## License
+
+Copyright (c) 2016 The OpenTracing Authors.
```
yurishkuro: `2016-present` would be more correct.

reimertz:
Do we really want to add a copyright to this file? Doesn’t that contradict our Apache license?

yurishkuro:
how does it contradict the license? The license is a permission to use the materials. Copyright exists even without declaring it, but stating it explicitly establishes who exactly it belongs to.

In the code repositories we always ask people to include headers with the source files that does the same ([example][1]), but since there are no headers in the doc files, having it here in the readme makes sense.

[1]: https://github.com/opentracing/opentracing-java/blob/master/opentracing-api/src/main/java/io/opentracing/Scope.java#L2

README.md
shresth27: @yurishkuro Done! Please review. Thanks. :)

PR #256, opened 2018-04-02
Propose a more precise tagline

bhs:
This is just one option of many... here's how it looks:

image

Curious what people think. cc @opentracing/otsc since not many people watch this repo.

### DIFF
--- a/_includes/hero.html
+++ b/_includes/hero.html
@@ -4,7 +4,7 @@
     {% include bash-window.html %}
 
     <section class="hero__copy">
-      <h1 class="bariol-thin">Vendor-neutral instrumentation and APIs for distributed tracing</h1>
+      <h1 class="bariol-thin">Vendor-neutral APIs and instrumentation for distributed tracing</h1>
       <a href="https://medium.com/opentracing/february-2018-opentracing-monthly-newsletter-7700560e9b04" target="_blank">February 2018: OpenTracing Project Update</a>
       <h2 class="dark-blue">Libraries available in {{ site.data.language.array.size }} languages</h2>
       <p class="dark-blue">
--- a/_includes/hero.html
+++ b/_includes/hero.html
@@ -4,7 +4,7 @@
     {% include bash-window.html %}
 
     <section class="hero__copy">
-      <h1 class="bariol-thin">Portable, general-purpose APIs for distributed tracing</h1>
+      <h1 class="bariol-thin">Vendor-neutral instrumentation and APIs for distributed tracing</h1>
       <a href="https://medium.com/opentracing/february-2018-opentracing-monthly-newsletter-7700560e9b04" target="_blank">February 2018: OpenTracing Project Update</a>
       <h2 class="dark-blue">Libraries available in {{ site.data.language.array.size }} languages</h2>
       <p class="dark-blue">
--- a/_includes/hero.html
+++ b/_includes/hero.html
@@ -4,8 +4,8 @@
     {% include bash-window.html %}
 
     <section class="hero__copy">
+      <h1 class="bariol-thin">Portable, general-purpose APIs for distributed tracing</h1>
       <a href="https://medium.com/opentracing/february-2018-opentracing-monthly-newsletter-7700560e9b04" target="_blank">February 2018: OpenTracing Project Update</a>
-      <h1 class="bariol-thin">A vendor-neutral open standard for distributed tracing.</h1>
       <h2 class="dark-blue">Libraries available in {{ site.data.language.array.size }} languages</h2>
       <p class="dark-blue">
       {% for language in site.data.language.array %}
--- a/_includes/hero.html
+++ b/_includes/hero.html
@@ -5,13 +5,15 @@
 
     <section class="hero__copy">
       <h1 class="bariol-thin">Vendor-neutral APIs and instrumentation for distributed tracing</h1>
-      <a href="https://medium.com/opentracing/february-2018-opentracing-monthly-newsletter-7700560e9b04" target="_blank">February 2018: OpenTracing Project Update</a>
+      <div class="libraries">
       <h2 class="dark-blue">Libraries available in {{ site.data.language.array.size }} languages</h2>
       <p class="dark-blue">
       {% for language in site.data.language.array %}
         <a href="{{language.url}}">{{ language.name }}</a>{% if forloop.last == false %},{% endif %}
       {% endfor %}
       </p>
+      </div>
+      <a href="https://medium.com/opentracing/february-2018-opentracing-monthly-newsletter-7700560e9b04" target="_blank">February 2018: OpenTracing Project Update</a>
     </section>
   </div>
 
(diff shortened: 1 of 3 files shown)


adriancole:
I don't want to back and forth on this, but I'd like to request to take Zipkin off opentracing's homepage. I'd suggest Jaeger instead.

There are a few reasons for this:

* Jaeger are better staffed and more inclined to want to support people starting with opentracing. OTOH, zipkin are except me unstaffed and being placed here only adds to our burden
* There's a large amount of confusion about the relationship between OpenTracing and Zipkin, with a fair amount of people thinking Zipkin is an OpenTracing project. Being perpetually on the front page only adds to this.
* As mentioned in various ways jaeger is "opentracing native" and so it makes a lot more sense for it to be on the home page.

I think this is just a legacy from before jaeger had a docker image. Anyway seems while you are adjusting your website you could consider addressing this.

JStickler:
Since OpenTracing is a standard, would it make sense to call that out in the tagline?

"Vendor neutral, standardized APIs for distributed tracing."

wu-sheng:
I dont think, exist realy vendor neutral. Stardard group members come from vendors. Even Apache projects can't be considered as vendor neutral. It is just open , welcome and high valued everyone.

saloni-garg:
Will the word **'Open-Source'** make the tagline more effective?

wu-sheng:
> Will the word 'Open-Source' make the tagline more effective?

@saloni-garg I think open source is obvious, because OpenTracing hosted in GitHub, and CNCF. Which both required the project open source, especially CNCF asked for Apache 2.0 license.

wu-sheng:
+1 about using Jaeger at home page. More opentracing native supported.

yurishkuro:
@bhs I think "general purpose" qualifier is so broad that it would invite similar criticism as before. I like the following qualifiers:
  * vendor-neutral (portable is similar, but more vague)
  * standardized (+1 to @JStickler)
  * cross-language / multi-language / polyglot (may not be necessary since we mention 9 langs next)
  * instrumentation APIs, not just APIs (unless we're planning to extend the mission)

victorNewRelic:
+1 to Yuri's suggestion to add the word instrumentation to result in something like:
> Portable, general purpose **instrumentation** APIs for distributed tracing.

tedsuo:
There is some concern with sing the term "Standard" on the website. Even though it's natural to say "standard tracing API" and "standard tags," the CNCF is not an official standards body. We can say "de-facto standard" or something similar – but it would be best if we avoided the word "Standard" in the main tagline until there is some resolution there.

We do strive for vendor-neutrality, and go to great length to make sure the API works well with all the existing vendors. If it isn't working well for someone then we want to fix it. So that is an intentional goal of the project and I think we should stick with it.

tedsuo:
Sorry you don't want Zipkin listed @adriancole! But I understand, and we can take it off the marquee. I would still like to list it under "supported tracers" as there are a number of ot-zipkin bridges available to people. Is that all right with you?

yurishkuro:
> There is some concern with sing the term "Standard" on the website ... the CNCF is not an official standards body.

meh... I wouldn't be hung up on this. When a large group of entities comes up with a "Technical specifications contained in a document that lays characteristics of a product such as levels of quality, performance, safety, or dimensions" (GATT definition), they end up with a standard, whether the group chooses to call itself a "standards body" or not.

Regardless, the current proposal is to use the word "standardized", which is different from "standard".

AloisReitbauer:
I was one of people objecting the word standard, because it simply isn't a standard. It might become one, but standards usually have very strict guidelines how they are defined, how implementations are tested, etc. 

I would be fine with an  "Open Source project aiming to define a standardized API for describing distributed traces".

adriancole:
I'm not sure who's actually supporting the zipkin tracers, certainly very
few folks from opentracing are, and issues go unresolved that affect us*.
However, it is more important to me to not have the marquee. So anyway,
sure.. let's just start with the marquee as the "supported" vs "things
being available" issue is a bit separate

* a couple issues that have gone no-where that limit us. The idea of
"support" seems to be in conflict when we are stuck due to things not
moving, yet have demand inflicted on us.
https://github.com/opentracing/opentracing-java/pull/212
https://github.com/opentracing/opentracing-java/issues/256

bhs:
@adriancole 

> I'd like to request to take Zipkin off opentracing's homepage

As you wish, sure. I spun this off as https://github.com/opentracing/opentracing.io/issues/257 since this PR is meant to be about the big-font copy which can/should be clearer.

bhs:
OpenTracing is valuable both for the decoupled "whitebox" application-level instrumentation and – now that the project is further along – the (mostly contributed) instrumentation of 3rd-party OSS out in the wild. As such, I wonder if we should go with something like `Vendor-neutral instrumentation and APIs for distributed tracing`. Screenshot:

image

As for the debate around the word "standard": with a "lower-case s" that word does _not_ imply – at least to my brain – that OpenTracing is a "standard" in the same way that a centimeter is a standard. I read that word as a synonym for "general-purpose". I removed it in this pass since I don't think the word "standard" is literally necessary and – empirically – it ruffles feathers.

yurishkuro:
> Vendor-neutral instrumentation and APIs for distributed tracing

+1. Only maybe APIs should come first, "APIs and instrumentation"?

bhs:
@yurishkuro 

> maybe APIs should come first

I agree... I'm embarrassed to admit this, but I started that way and got annoyed by the line breaks and couldn't figure out where the CSS was to control that particular `h1`! (It's not well-marked) Anyway, I found it. :)

image

@yurishkuro I've never loved the "vendor" aspect of "vendor-neutral", as most people use OpenTracing with an OSS tracing system (rarely referred to as a "vendor"). But "implementation" feels too wonky. I don't have a better suggestion than "vendor" so I'm ok with this, but if there are other ideas, now's a good time to suggest them.

AloisReitbauer:
You can use tool instead of vendor

tiffon:
An alternative to "vendor-neutral" might be "implementation agnostic"

bhs:
We could say "tool- and vendor-neutral"?

yurishkuro:
I think "vendor-neutral" is a well understood & intuitive enough term to not confuse people. 

Cf. https://www.gartner.com/it-glossary/vendor-neutral

bhs:
@yurishkuro I slept on it and agree that adding the "tool-" terminology doesn't make it clearer.

I also updated the css a bit for various reasons and fixed an issue with "medium" screen sizes while I was in the area.

image

adriancole:
I added https://github.com/opentracing/opentracing.io/issues/258 and waiting for others in openzipkin who have burden supporting tracers agree that zipkin should be removed entirely. It isn't up to people who don't contribute to make this decision

bhs:
This PR doesn't seem to be generating new opinions. I think what's here is a net improvement for sure... the one phrase that's both (a) still present and (b) somewhat problematic is "vendor-neutral"... I have done a bunch of googling and see that term used for numerous efforts where "vendors" (sic) are often OSS. As such, I think the usage here is still idiomatic (as Yuri pointed out [above](https://github.com/opentracing/opentracing.io/pull/256#issuecomment-378443256)).

I'd like to merge this before Monday AM if there are no strong objections.

Issue #265, opened 2018-04-15
The skywalking.org site (which we link to) has been hacked


tiffon:
Looks like http://skywalking.org/ has been hacked.

For instance, it just redirected me to: http://www.memevibe.com/2017/05/01/top-10-handsome-men-usa/

We should probably update the link to point refer http://skywalking.apache.org/

wu-sheng:
Yes, agree. We should update the website.

wu-sheng:
PR submitted.

wu-sheng:
Merged and fixed

PR #266, opened 2018-04-15
Update tracers.yaml

wu-sheng:
Update url of SkyWalking project based on @tiffon 's #265 

`http://skywalking.org` is no longer used by SkyWalking project. We have moved into 
http://skywalking.apache.org/ , and keep using http://skywalking.io


reimertz:
I think we should merge this ASAP. I'll merge this if no one has objections to why we shouldn't do it.

PR #267, opened 2018-06-03
Bugfix and document local setup procedure

tedsuo:
Walking through a clean local install, I noticed a couple of bugs.

* submodules were not installing correctly on `make setup`
* `make build` was failing due to theme dependencies not being setup.

I fixed these bugs, and I added local setup instructions to the README. I also removed the outdated CONTRIBUTING page; going forwards I'd prefer this info to be directly in the README. The submodules were pointing at outdated git commits, which were automatically bumped via `submodule --init`, so I checked those in as well.

Please double check that these were the "correct" fixes to make, and that the new instructions in the README work for you as well as me.


tedsuo:
@lucperkins I'm going to pull this in just to keep moving, but I'd still love your feedback.

PR #295, opened 2018-06-11
Copy/paste legacy content into v2.0 framework.

JStickler:
Copy/paste legacy content into v2.0 framework in preparation for the docathon this week.

I added a few pages that were missing from v2.0, but basically kept the same file names and organization as the v1.0 site.   I think I've got everything, but I'll give it a second pass tomorrow just to make sure.

### DIFF
--- a/content/documentation/api.md
+++ b/content/documentation/api.md
@@ -1,4 +1,9 @@
 ---
 title: API
 ---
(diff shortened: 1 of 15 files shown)


tedsuo:
Awesome! Thanks @JStickler!

PR #297, opened 2018-06-12
Install note about proper Hugo version

burningion:
There's a newer release of Hugo that breaks docs installation for me on Mac OS, .42.

I get an error `panic: AllThemes not set`, and the `make` command fails.

As a temporary workaround, adding a note about the proper version, and how to install it on Mac OS with Homebrew.


tedsuo:
Thanks @burningion.

PR #301, opened 2018-06-13
Write golang quick start documentation

black-adder:
Signed-off-by: Won Jun Jang <wjang@uber.com>


tedsuo:
Thanks @black-adder looks good!

PR #303, opened 2018-06-13
Adding Katacoda topic with links to courses.

JStickler:


tedsuo:
Nice! So excited for Katacoda, what a great resource.

PR #311, opened 2018-06-14
Add content for Java's Tracers.

carlosalberto:


content/guides/java/tracers.md
```diff
@@ -2,7 +2,67 @@
...
+
+Different **Tracer** implementations vary in how and what parameters they receive at initialization time, such as:
+
+* Component name for this application's traces.
+* Tracing endpoint.
+* Tracing credentials.
+* Sampling strategy.
+
+
+For example, initializing the **Tracer** implementation of `Jaeger` looks like this:
```
jpkrohling: s/looks like this/might look like this/

The "might" is because it could be also as simple as `Configuration.fromEnv().getTracer()`

content/guides/java/tracers.md
```diff
@@ -2,7 +2,67 @@
...
+
+  SamplerConfiguration samplerConfig = new SamplerConfiguration("const", 1);
+  ReporterConfiguration reporterConfig = new ReporterConfiguration(true, null, null, null, null);
+  Configuration config = new Configuration(service, samplerConfig, reporterConfig);
+
+  // Get the actual OpenTracing-compatible Tracer.
+  Tracer tracer = config.getTracer();
+```
+
+One a **Tracer** instance is obtained, it can be used to manually create **Span**, or pass it to existing instrumentation for frameworks and libraries:
```
jpkrohling: s/One/Once/

content/guides/java/tracers.md
```diff
@@ -2,7 +2,67 @@
...
+
+```java
+  // OpenTracing Redis client. It can be *any* OT-compatible tracer.
+  Tracer tracer = ...;
+  new TracingJedis(tracer);
```
jpkrohling: Nice name, I had to double check this was the real name, not a typo :D

content/guides/java/tracers.md
```diff
@@ -2,7 +2,67 @@
...
+```
+
+#### Global Tracer
+
+In order to not force the user to keep around a **Tracer**, the **io.opentracing.util** artifact includes a helper **GlobalTracer** class implementing the **io.opentracing.Tracer**, which, as the name implies, acts as as a global instance that can be used from anywhere. It works by forwarding all operations to another underlying **Tracer**, that will get registered at some future point.
```
jpkrohling: Might be better to state the actual artifact coordinates using the standard `g:a[:v]` notation: `io.opentracing:opentracing-util`, possibly with a link to the git repository?

content/guides/java/tracers.md
```diff
@@ -2,7 +2,67 @@
...
+  // using your CustomTracer.
+  Span span = GlobalTracer.get().buildSpan("foo").start();
+```
+
+#### Using Tracer specific features
+
+For using **Tracer**-specific features, the instance needs to be casted back to the original type:
+
+```java
+  ((CustomTracer)tracer).customFeature(100);
```
jpkrohling: I wonder if we should add a `unwrap(Class<Tracer>)` method to the interface, to avoid this cast here... Similar to what [JPA has](https://docs.oracle.com/javaee/6/api/javax/persistence/EntityManager.html#unwrap(java.lang.Class)).

content/guides/java/tracers.md
```diff
@@ -2,7 +2,67 @@
 title: "Java: Tracers"
 ---
 
-* Setting up your tracer
```
jpkrohling: How about a small intro here? Something like:

> A Tracer is the actual implementation that will record the spans and publish them somewhere. How your application handles the actual Tracer is up to you: either consume it directly throughout your application or store it in the GlobalTracer for easier usage with instrumented frameworks.

content/guides/java/tracers.md
```diff
@@ -2,7 +2,67 @@
...
+```java
+  import io.opentracing.Tracer;
+
+  SamplerConfiguration samplerConfig = new SamplerConfiguration("const", 1);
+  ReporterConfiguration reporterConfig = new ReporterConfiguration(true, null, null, null, null);
+  Configuration config = new Configuration(service, samplerConfig, reporterConfig);
+
+  // Get the actual OpenTracing-compatible Tracer.
+  Tracer tracer = config.getTracer();
+```
```
jpkrohling: It would be worth mentioning the Tracer Resolver as well. Something like:

> If your tracer supports it, the [`TracerResolver`](https://github.com/opentracing-contrib/java-tracerresolver) can also be used. With this approach, there's no tracer-specific initialization code in your application: `Tracer tracer = TracerResolver.resolveTracer();`

content/guides/java/tracers.md
```diff
@@ -2,7 +2,67 @@
 title: "Java: Tracers"
 ---
 
-* Setting up your tracer
-* Using the Global Tracer
-* Accessing the Active Span
```
jpkrohling: I guess it's missing this item: "Accessing the Active Span"

content/guides/java/tracers.md
carlosalberto: > might look like this

:) Makes sense, definitely.

content/guides/java/tracers.md
carlosalberto: Damn, thanks, will fix ;)

content/guides/java/tracers.md
carlosalberto: I covered this in the `Intro` section. Think we should put them here as well?

content/guides/java/tracers.md
carlosalberto: Oh, yes, I wondered about the same. Sounds like we should open an Issue in the java repo to keep track of this potential improvement?

content/guides/java/tracers.md
carlosalberto: Makes sense, will add it.

content/guides/java/tracers.md
carlosalberto: Oh, yes, had forgotten about it ;)

content/guides/java/tracers.md
carlosalberto: I covered that in the `Scopes` section (scopes.md, that is)

carlosalberto:
Updated ;)

content/guides/java/tracers.md
jpkrohling: https://github.com/opentracing/opentracing-java/issues/282

jpkrohling:
Looks great! I'll let @tedsuo or another native speaker proofread/merge.

tedsuo:
Thanks!

Issue #305, opened 2018-06-13
Python: Quickstart 


burningion:
Taking from the Java quickstart: 

* Link to Python walkthroughs / tutorials
* Setting up your tracer
* Start a Trace
* Create a Child Span
* Make An HTTP request
* View your trace

tedsuo:
Site is launched!

PR #313, opened 2018-06-14
Add Content: Python Quickstart

burningion:
This is a WIP, trying to flesh out a Python Quickstart to address issue #305. 

Open to feedback about structure.

For now, the great [Python OpenTracing tutorial](https://github.com/yurishkuro/opentracing-tutorial/tree/master/python) by Yuri seems to have a lot of overlap with what this should be? So for now, I've linked to that, and tried to start from an iPython interactive shell to send off first spans.

Let me know if this makes sense, or if we want to change the Quickstart structure to be substantially different from what I've got here.

Besides this, I'd like to add in some images of the Jaeger web interface if nobody has any problems with that. I could also do other vendors too, if there are stronger feelings.

Happy to hear feedback, and thanks for all the work team!

### DIFF
--- a/content/guides/python/quickstart.md
+++ b/content/guides/python/quickstart.md
@@ -2,18 +2,120 @@
 title: "Python: Quick Start"
 ---
 
-# Installation
+# Installing an OpenTracing Platform
 
-First, start by spinning up an instance of a tracing backend. In this quick start, we'll use Jaeger:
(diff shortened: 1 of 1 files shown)


tedsuo:
Thanks @burningion. This looks like it is in the right direction. In general, we are going to use Jaeger for quickstart examples. We'd like to come up with a way to emphasize that you can select different tracers, but we can decide that later, across all the guides.

PR #314, opened 2018-06-15
Add Content: Python Quickstart

burningion:
This continues the Work in Progress that was already merged in. Adds images to the walkthrough, and hopefully goes into a bit more depth.

Still want to put more details in to #305, especially for viewing traces with tags, and how that helps when debugging.

Will probably add a final check when viewing for whether the final exercise of web requests either failed or were successful.


tedsuo:
Thanks @burningion. Some notes:

* We'll do a pass cleaning up language and phrasing later, as we want to apply the same structure and terminology across all the language guides. So I'm not worried about copy editing right now.
* For errors, it would be great to have them pulled into their own section, and use the standard error tags & logs to explain this process to users.: 
  * https://github.com/opentracing/specification/blob/master/semantic_conventions.md#captured-errors
  * tag with "error" => `true`
  * log "event" => "error" and "error.object" => `exception`
* Add a section on Inject/Extract from HTTP Headers.
  * It would be good to make a "two-hop" example that involved injecting and extracting span context in an http header. There should be an explanation that you would then move this code into a library/plugin (because it's generic) and then share it in OT-contrib. 
* Once the new Python API is released, we should update this guide to use the ActiveSpan concept.

Happy to discuss. In the meantime, I'll go ahead and merge this.

Issue #272, opened 2018-06-04
Basics: Tracers


tedsuo:
* Choosing a Tracer
* Installing a Global Tracer

PikBot:
@tedsuo, what would be the difference in this section and the "Supported Tracers" section?

JStickler:
I would think that this section would link to the list of supported tracers, but also include some guidance on what to look for in a tracer, how different tracers might meet different business needs, etc.  Also instructions for how to install a global tracer (https://github.com/opentracing/opentracing-java/blob/master/opentracing-util/src/main/java/io/opentracing/util/GlobalTracer.java, https://github.com/opentracing/opentracing-go/blob/master/globaltracer.go) as well as more information about what a global tracer is, why you need one, etc.

PR #315, opened 2018-06-20
Add docs for Tracers

PikBot:
Signed-off-by: Prakriti <prakritibansal98@gmail.com>

refs #272


content/docs/overview/tracers.md
```diff
@@ -1,3 +1,63 @@
...
+The sampling rate of the tracer can either be fixed or dynamic. Dynamic sampling rate supports rate limiting and probabilistic sampling strategies whereas fixed sampling only supports probabilistic sampling strategies. Dynamic sampling rates can adapt to changes in the network behaviour. Jaeger supports dynamic sampling.
+
+<br>
+## Global tracers
+The first step in instrumenting your application with OT specification is registering the tracer with GlobalTracer. This step basically indicates the tracer’s endpoint and the service that will send data to the server. The GlobalTracer forwards all the methods to the tracer that has been configured. The code looks as follows (the tracer is Jaeger):
+
+
+
+
+```java
```
tedsuo: For the overview sections, we should avoid code examples. Those can go in the individual language guides.

content/docs/overview/tracers.md
```diff
@@ -1,3 +1,63 @@
...
+It is usually better to choose a tracer with better OT support and a larger diversity of OT-compatible clients for various programming languages. Jaeger is highly OT compatible and has clients available in most of the programming languages.
+
+#### Storage support
+The chosen tracer must be backed by high performance storage support. For production application, Elasticsearch or Cassandra are the most suitable options. MySQL, on the other hand, is not recommended due to performance issues. If persistence is not required, in-memory can also be used by the tracer.
+
+#### Sampling Rate
+The sampling rate of the tracer can either be fixed or dynamic. Dynamic sampling rate supports rate limiting and probabilistic sampling strategies whereas fixed sampling only supports probabilistic sampling strategies. Dynamic sampling rates can adapt to changes in the network behaviour. Jaeger supports dynamic sampling.
+
+<br>
+## Global tracers
```
tedsuo: This could be expanded into a "setting up your tracer" section.

content/docs/overview/tracers.md
```diff
@@ -1,3 +1,63 @@
...
+
+<br>
+
+## Choosing a Tracer
+It is important that a tracer that meets your tracing requirements as well as fits the needs of your application is chosen. This section discusses some of the parameters for choosing a tracer.
+#### OT compatibility
+The tracer clients must be compatible with the OT API, in order to be used with OpenTracing API instrumentation. It is required that all the OT compatible tracers provide the implementation of tracer interface.
+
+#### OT compatible clients
+It is usually better to choose a tracer with better OT support and a larger diversity of OT-compatible clients for various programming languages. Jaeger is highly OT compatible and has clients available in most of the programming languages.
```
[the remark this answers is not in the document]
PikBot: Yeah, actually I was pretty confused about what to add under this, is there something I could add?

content/docs/overview/tracers.md
PikBot: Would this just be a section which goes through installation of a tracer like Jaeger or Zipkin through docker image?

content/docs/overview/tracers.md
```diff
@@ -1,3 +1,63 @@
 ---
 title: Tracers
 ---
+
+OpenTracing provides an open, vendor-neutral standard for distributed tracing systems instrumentation, providing an interface to the different distributed tracers. This means that instrumentation would remain the same irrespective of the tracer system being used by the developer. In order to instrument the web application using OpenTracing specification, a compatible OpenTracing tracer must be deployed and listening for incoming span requests. A list of the all the OpenTracing tracers is available [here](./../../supported-tracers). The following section discusses Global Tracers and the various tracer properties that must be taken into account while choosing one.
+
+<br>
```
yurishkuro: random markup is not needed

content/docs/overview/tracers.md
```diff
@@ -1,3 +1,63 @@
...
+
+## Choosing a Tracer
```
yurishkuro: I am not sure this is the right way to phrase the question. Tracer is just an extension of the tracing backend, one cannot choose them independently. And picking the backend purely based on the features of the tracer (instrumentation library) is not useful. 

I think this page can be simply a list of tracing systems that supply OT-compatible tracers. Performing a feature comparison of those tracing systems is not in scope of the documentation, since it's going to become out of date quickly. For example:

The following tracing backends have OpenTracing-compatible tracing libraries:

| Tracing System | Go tracer | Java tracer  | 
| ------------- |-------------| -----|
| name      | [link](link) | n/a |

Given that we probably have more languages that tracing systems, may want to transpose the table. Or not do a table, since it's going to be hard to edit, just have a list of tracing backends with sublists of clients by language.

tedsuo:
In general, I wold say the "Tracer" section needs to explain the following:

* The Tracer Interface and Tracer Implentations
* Setting up a Tracer
  * Initialize your tracer at application startup
  * Set the tracer as the Global Tracer
* Starting a new Trace
* Accessing the Active Span
* Propagating a Trace with Inject and Extract

Basically, the Tracer Overview should be a cleaned up version of the Tracer section of the specification:
https://github.com/opentracing/specification/blob/master/specification.md#tracer

PR #318, opened 2018-07-02
Add Python inject-extract docs

PikBot:


tedsuo:
Looks like the Java docs, so I'm going to accept.

PR #319, opened 2018-07-07
Add Python: Tracer docs

PikBot:
Signed-off-by: Prakriti <prakritibansal98@gmail.com>

### DIFF
--- a/content/guides/python/tracers.md
+++ b/content/guides/python/tracers.md
@@ -1,0 +1,49 @@
+---
+title: "Python: Tracers"
+---
+
+## Setting up your Tracer
+A `Tracer` is the actual implementation that will record the `Spans` and publish them somewhere. Python `Tracer` implementation defines both the public Tracer API and provides a default no-op behavior.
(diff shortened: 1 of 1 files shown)


content/guides/python/tracers.md
```diff
@@ -0,0 +1,97 @@
...
+    scope.span.set_tag('...', '...')
+```
+
+### Propagating a Trace with Inject/Extract
+In order to trace across process boundaries in distributed systems, services need to be able to continue the trace injected by the client that sent each request. OpenTracing allows this to happen by providing inject and extract methods that encode a span's context into a carrier.
+The `inject` method allows for the `SpanContext` to be passed on to a carrier. The `extract` method does the exact opposite. It extracts the `SpanContext` from the carrier.
+
+Implemented examples of inject/extract and details can be found [here](./inject-extract).
+
+![Trace Propagation](/img/overview:tracers/Extract.png)
```
tedsuo: can you add a basic example using http here? The Inject/Extract docs should go into all the options, but a basic example would be good to have.

tedsuo:
Thanks!

PR #320, opened 2018-07-07
Add Overview: Scopes and Threading docs

PikBot:


content/docs/overview/scopes-and-threading.md
```diff
@@ -0,0 +1,42 @@
+---
+title: "Scopes and Threading"
+---
+## Introduction
+
+The OpenTracing API allows for only one `span` in a thread to be active at any point in time. There can be other spans involved with the same thread, which satisfy the following conditions:
```
tedsuo: It would be better to start with an introduction to the Active Span concept, then mention scopes, then get into these details.

content/docs/overview/scopes-and-threading.md
```diff
@@ -0,0 +1,42 @@
...
+As it is inconvenient to pass an active `Span` from function to function manually, so OpenTracing requires that every `Tracer` contain a `ScopeManager`. A `ScopeManager` grants access to the active Span through a `Scope`. This means that a developer can access any active `Span` through a `Scope`.
+
+## Moving a span between threads
+Using the OpenTracing API, a developer can transfer the spans among different threads. A `Span`’s lifetime might start in one thread and end in another. Passing of scopes to another thread or callback is not supported.
+The internal timing breakdown of a `Span` might look like as shown below.
+```
+[ ServiceHandlerSpan                                 ]
+|·FunctionA·|·····waiting on an RPC······|·FunctionB·|
+
+--------------------------------------------------> time
```
tedsuo: It's not clear how this example is relevant to moving a span between threads.

content/docs/overview/scopes-and-threading.md
```diff
@@ -0,0 +1,42 @@
...
+```
+
+The “ServiceHandlerSpan” is active while it’s running in FunctionA and FunctionB, but remains inactive while waiting on an Remote Procedure Call (RPC). The RPC will presumably have a `Span` of its own, however, we are only concerned with the propagation of “ServiceHandlerSpan” from FunctionA to FunctionB.
+Using the `ScopeManager` API it is possible to fetch the `span()` in FunctionA and re-activate it in FunctionB after RPC terminates. The steps are as below:
+
+1. Start a Span via either `startManual` or `startActive(false)` to prevent the `Span` from being finished upon `Scope` deactivation.
+
+2. Invoke `tracer.scopeManager().activate(span, false)` to re-activate the `Span` and get a new `Scope`, then `deactivate()` it when the `Span` is no longer active.
```
tedsuo: Many of these details feel Java-specific, and reference the Java API. For the overview, we want to have a more generic way of discussing these concepts.

tedsuo:
@PikBot looks great! I'm going to merge it.

PR #353, opened 2018-10-19
Add Jaeger tracers

yurishkuro:
Signed-off-by: Yuri Shkuro <ys@uber.com>


austinlparker:
These are missing `otVersion` tags, could you add them? (or just set it to `otVersion: latest`)

yurishkuro:
Sorry, didn't see the comment. I don't want to keep maintaining those version manually. I don't think they add a lot of value.

PR #357, opened 2018-10-23
Add AppOptics tracer to the OpenTracing Registry

cce:
adds a mention of the OT tracer in https://github.com/appoptics/appoptics-apm-go


austinlparker:
Thanks for the submission!

PR #364, opened 2018-11-05
Add go-gin to the registry

stuart-mclaren:


stuart-mclaren:
Thanks for your review Austin.
Is there a way to propose this moves to https://github.com/opentracing-contrib?

austinlparker:
Sure, we'd be thrilled to have it. Let me check and see what our process for this is and I'll open an issue on your repo to track it.

stuart-mclaren:
Thanks!

PR #369, opened 2018-11-20
Fix bashWindowText to match change in opentracing-contrib/examples

jcreedcmu:
Since
https://github.com/opentracing-contrib/examples/commit/dd9b3e3c349748805c619c7950938345d240e51e
landed, the go example mentioned in `bashWindowText` produces jaeger.thrift instead
of zipkin.thrift, and consequently hits udp port 6831 instead of udp port 5775.

Not sure if `content/docs/getting-started/_index.md` also needs to change.


jcreedcmu:
(at least I observed the example to work successfully after I made this change, and not before. Apologies if I'm mistaken and missing something obvious...)

PR #370, opened 2018-11-27
Add couple of ruby projects to the registry

indrekj:


austinlparker:
lgtm, thanks for the contribution!

PR #374, opened 2018-12-06
Minor fixes to ease first time contributor life

nimeshksingh:
1. Have `make get-spec-docs` initialize submodules.
2. Delete and .gitignore the resources directory.


nimeshksingh:
Yeah, I discovered the same sad fact.  I think I'll still update the netlify version and commit the resources files that match that version.

PR #378, opened 2018-12-12
add java-sqs instrumentation library to registry

jmhofer:


austinlparker:
Thanks for the contribution!

PR #383, opened 2018-12-20
fix escape in the link to Jaeger

autopp:
Hi, I found the broken link to Jaeger in https://opentracing.io/docs/supported-tracers/

Before:
image

After:
image

This seems to be a problem of hugo, but the cause has not been examined yet, sorry.


yurishkuro:
Thanks!

autopp:
Thanks for merge!

PR #386, opened 2019-01-04
add lua to supported languages

austinlparker:


yurishkuro:
I have a few Qs before we merge:
- has anyone from OTSC reviewed the api?
- does the repository have official maintainers? Are they listed somewhere?
- which tracers implement the API?
- where has it been used in instrumentation? Any public repos?

austinlparker:
There's a comment thread with Ted involved on that repo - https://github.com/opentracing/opentracing-lua/pull/2

I had assumed that since it was in the OT organization things were copacetic. Not sure if there's any public consumers of it, though.

PR #388, opened 2019-01-08
Add Python Requests instrumentation

rmfitzpatrick:
Includes registry entry for https://github.com/opentracing-contrib/python-requests


austinlparker:
Thanks for these!

PR #392, opened 2019-01-09
Fix broken link on python/tracers

iandelahorne:
The link for inject-extract on guides/python/tracers was 404'ing.
Link this to the correct path


austinlparker:
Nice catch, thanks for the fix!

PR #406, opened 2019-02-01
update registry entry for Datadog .NET Tracer

lucaspimentel:
- add supported OpenTracing version
- fix incorrect license
- rename tag for consistency


austinlparker:
Thanks!

PR #408, opened 2019-02-05
Add contribution: opentracing-basic lib for java

mcumings:
Adds reference to the opentracing-basic lib here: https://github.com/eBay/opentracing-basic/

### DIFF
--- a/content/registry/java-opentracing-basic.md
+++ b/content/registry/java-opentracing-basic.md
@@ -6,7 +6,7 @@ tags:
   - java
 repo: https://github.com/eBay/opentracing-basic/
 license: Apache 2.0
-description: Basic OpenTracing bridge/stub library for Java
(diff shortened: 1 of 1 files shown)


content/registry/java-opentracing-basic.md
```diff
@@ -0,0 +1,12 @@
+---
+title: opentracing-basic
+registryType: tracer
+tags:
+  - opentracing
+  - java
+repo: https://github.com/eBay/opentracing-basic/
+license: Apache 2.0
+description: Basic OpenTracing bridge/stub library for Java
```
yurishkuro: I would suggest to take the wording from the repo itself, which is more explanatory that this description:

> The goal of this library is to attempt to implement the generic semantics of the OpenTracing API and allow for the main implementation details to be provided into the library by the consumer.

mcumings: Sure thing.  The other registrations were quite brief so I didn't want to go overboard.

yurishkuro: many other entries don't have such a generic name

PR #424, opened 2019-07-06
Add Python Asyncpg intstrumentation

condorcet:
Opentracing instrumentation for asyncpg https://github.com/magicStack/asyncpg library

### DIFF
--- a/content/registry/python-asyncpg.md
+++ b/content/registry/python-asyncpg.md
@@ -4,6 +4,7 @@ registryType: instrumentation
 tags:
     - python
     - opentracing
+    - asyncpg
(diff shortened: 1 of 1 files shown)


content/registry/python-asyncpg.md
```diff
@@ -0,0 +1,12 @@
+---
+title: asyncpg-opentracing
+registryType: instrumentation
+tags:
+    - python
+    - opentracing
```
yurishkuro: I recommend adding asyncpg as a tag, so that it can be searched

condorcet: @yurishkuro Thank you. Done.

PR #441, opened 2020-07-08
instrumentation of newest go-redis

lyb0307:
I searched for the redis client of golang with opentracing, they are all out of date. 
I noticed there was an instrumentation of go-redis (https://github.com/opentracing-contrib/goredis) supports the v6 version of go-redis. 
While the latest version is v8 and they have changed their "WrapProcess" method to "Hook" interface.
Also I noticed go-redis has their own tracing based on OpenTelemetry (https://redis.uptrace.dev/tracing/).
So I just modified it to support opentracing.
Here it is.

### DIFF
--- a/content/registry/go-opentracing-go-redis
+++ b/content/registry/go-opentracing-go-redis
@@ -6,7 +6,7 @@ tags:
     - redis
     - go-redis
 repo: https://github.com/lyb0307/opentracing-go-redis
-license: WTFPL
+license: MIT
 description: I searched for the redis client of golang with opentracing, they are all out of date. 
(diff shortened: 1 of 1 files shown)


content/registry/go-opentracing-go-redis
```diff
@@ -0,0 +1,18 @@
+---
+title: opentracing-go-redis
+registryType: instrumentation
+tags:
+    - golang
+    - redis
+    - go-redis
+repo: https://github.com/lyb0307/opentracing-go-redis
+license: WTFPL
```
yurishkuro, line 9: you may want to consider a different, more widely used license, like MIT or APLv2.

lyb0307: I have changed it to MIT. Thank you for the reply~

PR #443, opened 2020-08-17
Create javascript-foxx-tracer.md

adityamukho:
Most tracing libraries in the nodeverse are asynchronous, and so do not work in the synchronous V8 runtime that ArangoDB uses to run its Foxx services. _foxx-tracer_ bridges this gap by being a 100% synchronous, dedicated module built for the Foxx runtime.

### DIFF
--- a/content/registry/javascript-foxx-tracer.md
+++ b/content/registry/javascript-foxx-tracer.md
@@ -1,6 +1,6 @@
 ---
 title: foxx-tracer - An instrumentation library for Foxx Microservices
-registryType: <instrumentation/tracer>
+registryType: tracer
 tags:
     - javascript
     - foxx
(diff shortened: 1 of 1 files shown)


content/registry/javascript-foxx-tracer.md
```diff
@@ -0,0 +1,14 @@
+---
+title: foxx-tracer - An instrumentation library for Foxx Microservices
+registryType: <instrumentation/tracer>
```
yurishkuro, line 3: * `<>` syntax is invalid
* is this a tracer or the instrumentation?  A tracer is something that _implements_ OpenTracing API, instrumentation is something that _uses_ the API to describe transactions.

adityamukho: Sorry, this should be labelled a tracer. Sending an update shortly.

PR #445, opened 2020-11-23
adding Hypertrace project

dnielsen:
Hypertrace is an extensible distributed tracing project


pavolloffay:
Please add this to https://opentracing.io/docs/supported-tracers/. The registry tracks only instrumentations.

PR #456, opened 2021-08-10
Ignore Hugo-generated resources folder

chalin:
/cc @tedsuo


chalin:
@austinlparker - could you review this PR?

PR #462, opened 2022-04-25
Noindex & nofollow site pages

chalin:
- Addresses https://github.com/open-telemetry/opentelemetry.io/issues/1316

Preview:

- https://deploy-preview-462--opentracing.netlify.app, has no "noindex" tag value
- https://deploy-preview-462--opentracing.netlify.app/specification/, then View Source to see the robots meta tag with noindex.

/cc @austinlparker @yurishkuro @tedsuo @caniszczyk

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -1,7 +1,6 @@
 title: The OpenTracing project
 baseURL: /
 languageCode: en-us
-enableRobotsTXT: true
 
 pygmentsUseClasses: true
 pygmentsCodefences: true
--- a/config.yaml
+++ b/config.yaml
@@ -1,6 +1,7 @@
 title: The OpenTracing project
 baseURL: /
 languageCode: en-us
+enableRobotsTXT: true
 
 pygmentsUseClasses: true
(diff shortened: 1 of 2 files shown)


yurishkuro (commented):
Wouldn't this completely "erase" the website from Google, even if people search for opentracing?

austinlparker:
@yurishkuro I believe so. I can see there being some flaws there, as people do use OT for various reasons still -- although, there are pointers from OTel docs back to OT (https://opentelemetry.io/docs/migration/opentracing/).

chalin:
> Wouldn't this completely "erase" the website from Google, even if people search for opentracing?

Yes. Might that be too radical?

An alternative might be to not index the content of the site, but still index the homepage?

Of course, I'm open to any other suggestions.

yurishkuro:
> An alternative might be to not index the content of the site other than the homepage.

yeah, I would at least leave the home page indexed, but other pages seem fine to exclude (even the registry).

yurishkuro (approved):

